Issue #3, opened 2015-08-20
Panics on drop will cause double drops.


Stebalien:
IntoIter and ArrayVec need to be poisoned to prevent dropping before being drained; otherwise, if one of the elements panics on drop, it will be dropped twice. I've worked on a couple of solutions but they all have design tradeoffs so I figured I'd just let you fix it.

Drain doesn't have this issue because the ArrayVec's length is set to 0 when the Drain is initialized.

Stebalien:
Test cases (abusing rust's panic on drop -> abort).
```rust
#[test]
#[should_panic]
fn test_drop_panic() {
    struct DropPanic;

    impl Drop for DropPanic {
        fn drop(&mut self) {
            panic!("drop");
        }
    }

    let mut array = ArrayVec::<[DropPanic; 1]>::new();
    array.push(DropPanic);
}

#[test]
#[should_panic]
fn test_drop_panic_into_iter() {
    struct DropPanic;

    impl Drop for DropPanic {
        fn drop(&mut self) {
            panic!("drop");
        }
    }

    let mut array = ArrayVec::<[DropPanic; 1]>::new();
    array.push(DropPanic);
    array.into_iter();
}
```

bluss:
Thanks! I guess this won't be easy..

bluss:
For the lols, please read #5. Hey you said you would leave it for me to fix! But I do think you have the sane version, and I the *insane*..

bluss:
Fixed in version 0.3.10

bluss:
Great, now it's fixed in smallvec too.

PR #4, opened 2015-08-20
Don't drop twice on panic.

Stebalien:


Stebalien:
This is *a* fix for #3 but not necessarily the best fix.

bluss:
I'm working on a fix too :) Let's compare..

bluss:
I wonder which of these fixes are classified as the most fragile.

Thanks for submitting this PR by the way, made me understand this much better as a real panic safety bug.

bluss:
I think the sane solution is to go back to using the external NoDrop with its own destructor (like arrayvec pre 0.3.8). That works, but do you think that's an accident?

Stebalien:
> I think the sane solution is to go back to using the external NoDrop with its own destructor (like arrayvec pre 0.3.8).

That's what I did first. However, it makes the arrayvec one byte larger due to the drop flag...

bluss:
One byte is just for the Arrayvec of u8. It makes it a whole element larger. It's an ok tradeoff for now..

Stebalien:
Also, that doesn't fix the IntoIter case (unless you put a NoDrop around the ArrayVec inside the IntoIter).

bluss:
See PR bluss/arrayvec/pull/6, it covers into iter too

bluss:
It is probably the case that panic on drop rules are simply not well defined. I'd prefer this guard case if it wasn't for the weird bug.

PR #5, opened 2015-08-20
Avoid double drop on panic bug

### LINKED ISSUE #3: Panics on drop will cause double drops.

bluss:
We need to make sure the NoDrop enum's discriminant is set to Dropped,
even if we panic during drop of one of the elements.

Bug reported by @stebalien

Fixes #3


bluss:
Crazy solution dude, no way.

PR #6, opened 2015-08-20
Use NoDrop to fix panic safety issues

### LINKED ISSUE #3: Panics on drop will cause double drops.

bluss:
ArrayVec::drop was not panic safe — if there would be a panic during an
element's drop, the discriminant would never be set to Dropped, and the
array elements would potentially double drop.

Fix this by going back to the old NoDrop composition. The NoDrop struct
thas its own Drop impl, that will trigger too on panic during an element's
drop. This serves to make ArrayVec::drop panic safe.

Also tweak IntoIter::drop to make it panic safe: set inner ArrayVec's
length before dropping any elements.

Thank you to @Stebalien for reporting this bug and providing the
excellent testcases in this commit.

Using NoDrop expands ArrayVec to have two drop flags again, but this
is a temporary tradeoff, drop flags will eventually go away.

Fixes #3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ sudo: false
 matrix:
   include:
     - rust: stable
+    - rust: beta
     - rust: nightly
     - rust: nightly
       env:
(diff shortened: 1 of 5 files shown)


Stebalien:
LGTM

bluss:
Using feature `nodrop/no_drop_flag` (requires nightly) eliminates the nodrop drop flag, and should recover the space again.

bluss:
Thank you for helping with this!

PR #8, opened 2015-08-23
Expose nodrop state through its API

arcnmx:
When using `NoDrop` with `no_drop_flag`, there is no way to assert whether the inner value has been dropped or not. This PR exposes `is_alive` for containing structures that may also use `unsafe_no_drop_flag` themselves.

### DIFF
--- a/nodrop/src/lib.rs
+++ b/nodrop/src/lib.rs
@@ -10,7 +10,7 @@
 //!
 //!
 
-#![cfg_attr(feature="no_drop_flag", feature(unsafe_no_drop_flag))]
+#![cfg_attr(feature="no_drop_flag", feature(unsafe_no_drop_flag, filling_drop, core_intrinsics))]
 
 extern crate odds;
 
@@ -20,7 +20,7 @@ use std::ops::{Deref, DerefMut};
 use std::ptr;
 use std::mem;
 
-/// repr(u8) - Make sure the non-nullable pointer optimization does not occur!
+// repr(u8) - Make sure the non-nullable pointer optimization does not occur!
 #[repr(u8)]
 enum Flag<T> {
     Alive(T),
@@ -49,6 +49,39 @@ impl<T> NoDrop<T> {
         // skip Drop, so we don't even have to overwrite
         mem::forget(self);
         inner
+    }
+
+    /// Extract the inner value from a mutable reference.
+    ///
+    /// Once extracted, it is **not** safe to use this instance beyond letting it drop.
+    #[inline]
+    pub unsafe fn take(&mut self) -> T {
+        match mem::replace(&mut self.0, Flag::Dropped) {
+            Flag::Alive(inner) => inner,
+            _ => debug_assert_unreachable()
+        }
+    }
+
+    /// Determines whether the inner value is safe to use or dereference.
+    #[inline]
+    #[cfg(not(feature="no_drop_flag"))]
+    pub fn is_alive(&self) -> bool {
+        match self.0 {
+            Flag::Alive(..) => true,
+            Flag::Dropped => false
+        }
+    }
+
+    /// Determines whether the inner value is safe to use or dereference.
+    #[inline]
+    #[cfg(feature="no_drop_flag")]
+    pub fn is_alive(&self) -> bool {
+        use std::intrinsics::discriminant_value;
+        unsafe {
+            let dropped: Flag<()> = Flag::Dropped;
+            let discriminant = discriminant_value(&self.0);
+            discriminant != discriminant_value(&dropped) && discriminant as u8 != mem::POST_DROP_U8
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


nodrop/src/lib.rs
```diff
@@ -38,6 +38,14 @@ impl<T> NoDrop<T> {
         NoDrop(Flag::Alive(value))
     }
 
+    /// Create a new **NoDrop** without a value.
+    ///
+    /// It is not safe to use the returned value without first checking `is_alive()`.
+    #[inline]
+    pub unsafe fn dropped() -> NoDrop<T> {
```
bluss: Interesting. Does this method have any use?

arcnmx: That one, I don't actually have a concrete use case for. It just seemed a
bit more appropriate than `new(mem::uninitialized())` for lazy
instantiation. The main purpose of this pr is to allow the discriminant to
double as an observable drop flag. The rest is auxiliary, though I do like
take().

bluss: I'll find use for take too, I think. I'd prefer to skip the dropped constructor if we don't have a use for it yet. just YAGNI it :)

arcnmx: Fair enough! Removing...

arcnmx:
Hm, this is wrong. The discriminant actually returns POST_DROP_U64. Not even sure why considering it's repr(u8)..? But fixing...

bluss:
Weird. For which `T`? I suspect that the discriminant is expanded if there would be padding, I think I remember reading about that.

arcnmx:
You are correct! Seems like bizarre behaviour to me, but fixed.

arcnmx:
Haha, or not, clearly.

arcnmx:
`thread 'test_discriminants' panicked at 'assertion failed: `(left == right)` (left: `29`, right: `29`)', nodrop/src/lib.rs:146`

That's.... nice.

bluss:
> thread 'test_discriminants' panicked at 'assertion failed: `(left == right)` (left: `29`, right: `29`)', nodrop/src/lib.rs:146

Nothing says undefined behavior like this :-)

arcnmx:
Welp, the discriminant_value intrinsic does indeed generate a value with bounds, which of course llvm's optimizer can do whatever it wants with.

@bluss how do you feel about this:
```rust
	pub fn is_alive(&self) -> bool {
		if let Flag::Alive(..) = self.0 {
			true
		} else {
			false
		}
	}
```

It works in debug and release, but I'm not sure if it's technically **defined** to work as expected.

bluss:
Requires some research I guess.

arcnmx:
That's why I went with the discriminant intrinsic in the first place, I assumed it would give me whatever value was in the discriminant, well, indiscriminantly (and therefore its behaviour would be a bit more defined). This approach is alright, and does work, but I'm not sure that we can guarantee it's correct.

Another option is to simply `transmute::<_, &(u8, T)>(self).0`, which may actually be considered more well defined (and also works fine), but only because it relies on compiler implementation details on how it lays out the enum.

bluss:
I think the if-let code has a range load too [(code)](https://play.rust-lang.org/?gist=934b253f8894864c89cc&version=nightly) in debug mode, llvm output.

bluss:
So yeah, using match, if let, or discriminant_value are all impossible I think.

arcnmx:
Wonder if we could just blackbox'ify the discriminant_value's output... Or otherwise remove the range metadata (with some intrinsic like transmute?)...

The thing is, it does have a range load, *but* it uses the switch default fallback if it's not alive, so actually should work in all cases. Not well defined though, it's feasible though unlikely that an optimization would make it check for a `1` instead.

But yeah, transmuting to read the discriminant (with adjustements for alignment fanciness) would work, ugh...

bluss:
It does seem like filling drop is at odds with reading the discriminant and with this feature

arcnmx:
Mm, I find it kind of strange that `unsafe_no_drop_flag` still performs a filling drop. All this effort to remove a single drop flag that will disappear anyway as soon as drop flags get removed from the language D:

bluss:
I'm hopeful filling drop and drop flags will be removed soon.

I don't see any great way forward for this PR with this situation -- the only way would be a hacky way to read out the enum manually with a u8 pointer, but we don't even have any way to get its offset/location.

bluss:
I'll probably close this issue, what do you think? Rust doesn't really give us an option here, mostly due to filling drop.

By the way, don't miss [this tidbit](https://www.reddit.com/r/rust/comments/3irc91/crate_arrayvec_a_vector_backed_by_a_fixed_size/cuj43qq). That may be really useful.

arcnmx:
Yeah, may as well drop it. Hopefully drop flags will disappear soon!

arcnmx:
@bluss would it make sense to put together another PR for `take()`? It feels a bit nicer than using `ptr::read` to move data out.

bluss:
Yes!

PR #12, opened 2015-09-10
Add `io::Write` implementation akin to `Vec`

tbu-:


tbu-:
Probably not as straight-forward as the others today...

bluss:
Ok this looks like something we should have. There is a myriad traits you need to implement for a rust collection! o_O

I'd like to write it like the following in mind

- No need for an extra method
- Use default for `write_all`.
- Use newlines and no "unsafe minimization" to uncramp the style.

```rust
    fn write(&mut self, data: &[u8]) -> io::Result<usize> {
        unsafe {
            let len = self.len();
            let mut tail = slice::from_raw_parts_mut(self.get_unchecked_mut(len),
                                                     A::capacity() - len);
            let result = (&mut tail).write(data);
            if let Ok(written) = result {
                self.set_len(len + written);
            }
            result
        }
    }
```

If it doesn't compile, take it as a sketch.

tbu-:
Here you go! I just removed the history, I hope that's fine.

bluss:
great. more PRs coming or are we ready to publish a new version?

tbu-:
No more PRs coming in right now. I looked over the list of traits in `std` and didn't find further traits to implement (but I guess I would've missed `io::Write` if I hadn't specifically needed that).

Ready to publish a new version!

bluss:
Great

PR #13, opened 2015-09-11
Add `ArrayString`, which is to `ArrayVec` what `String` is to `Vec`

tbu-:


tbu-:
To easily check that nothing was lost during the rename you can use

    git diff HEAD^:src/lib.rs HEAD:src/array_vec.rs

once you checked out the branch.

tbu-:
Note that I actually use that type in my crate, it's not just implemented for fun. :)

bluss:
cool, I'll look closely at this. thanks for starting simple.

bluss:
This is a breaking change by moving some public types. Notice all the types and traits under `arrayvec::*` here: [docs](http://bluss.github.io/arrayvec/doc/arrayvec/index.html)

bluss:
Simpler to not move ArrayVec right now.

src/array_string.rs
```diff
@@ -0,0 +1,165 @@
...
+        self.vec.extend(&mut bytes);
+        assert!(bytes.next().is_none());
+        None
+    }
+
+
+
+    /// Make the string empty.
+    pub fn clear(&mut self) {
+        mem::replace(self, ArrayString::new());
```
bluss: replace isn't needed. Just `*self = ` would work, even better, just set_len(0).

tbu-:
@bluss Added your commit and move `ArrayVec` back.

src/array_string.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// let none2 = string.push_str("d");
+    /// let overflow2 = string.push_str("ef");
+    ///
+    /// assert_eq!(&string[..], "ad");
+    /// assert_eq!(none1, None);
+    /// assert_eq!(none2, None);
+    /// assert_eq!(overflow1, Some("bc"));
+    /// assert_eq!(overflow2, Some("ef"));
+    /// ```
+    pub fn push_str<'a>(&mut self, s: &'a str) -> Option<&'a str> {
```
bluss: A `&str` is a simple copyable type so we don't need to return it again here. Could just be bool.

Although, if we take your request for a more consistent interface, these methods should be introduced to be equivalent to the String method signatures? We don't have to worry about backward compat with new methods.

src/array_string.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// let none1 = string.push('a');
+    /// let none2 = string.push('b');
+    /// let overflow = string.push('c');
+    ///
+    /// assert_eq!(&string[..], "ab");
+    /// assert_eq!(none1, None);
+    /// assert_eq!(none2, None);
+    /// assert_eq!(overflow, Some('c'));
+    /// ```
+    pub fn push(&mut self, c: char) -> Option<char> {
```
bluss: Same, no need to return the char here.

src/array_string.rs
tbu-: On `String` these methods return nothing, so I'd suggest returning `Result<(), CapacityError>` here. That way the result becomes a `#[must_use]` and the user has to think about what to do in this case, or assert by writing `.unwrap()` that the capacity suffices.

bluss: This is ok, but it sort of abandons the idea of having an identical / drop in interface (previous issue raised for ArrayVec). Maybe this is saner.. the method names are the same, but you are suddenly forced to consider capacity overflow errors?

tbu-: I don't know, I consider silently failing as a really bad thing. Currently my code using `Array*` looks like `assert!(a.push(..).is_none());`, but if I forget this check, no warning will be issued. This would be different if `push` returned a result.

You can still use it as a "drop-in" for `String` if you really want to, and aren't afraid of seeing a few warnings (or silencing them).

bluss: Sounds good to me. We can do that for ArrayVec eventually too, but it's a bit harder since it's a breaking change.

bluss: If we use a Result anyway, maybe we should return the &str / char anyway for convenience?

tbu-:
@bluss Done.

bluss:
I'll look at this in the evening. I might merge and do some changes on top of that. The error design needs some thought.

bluss:
ArrayString is in arrayvec 0.3.12, took some time.

tbu-:
Cool!

PR #25, opened 2016-02-06
Support no_std in nodrop

gereeter:
See #20.

### DIFF
--- a/nodrop/src/lib.rs
+++ b/nodrop/src/lib.rs
@@ -17,13 +17,17 @@
 #![cfg_attr(feature="no_drop_flag", feature(unsafe_no_drop_flag))]
 #![cfg_attr(feature="use_needs_drop", feature(core_intrinsics))]
 
+#![cfg_attr(not(test), no_std)]
+#[cfg(test)]
+extern crate core;
+
 extern crate odds;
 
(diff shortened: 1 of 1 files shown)


gereeter:
Ah - you are testing against Rust 1.2, which doesn't have stable `core`. Should I put this behind a cargo feature?

bluss:
Right, we need a solution for that. Nice crate that doesn't need anything from std.

I want to use a `std` feature and have it enabled by default. There was some discussion about what the community's standard solution for this was going to be, but I think that's fine.

Moving the baseline from Rust 1.2 is fine, but not all the way up to Rust 1.6 yet I think?

Discussion in https://github.com/rust-lang-nursery/libc/pull/146

bluss:
Heck, I think Rust 1.6 is fine, we can do a bump to version 0.2. This crate doesn't take any harm from 0.1 -> 0.2 bumping does it, it doesn't share types really with other crates, so the implications should be minimal.

bluss:
To edit that thought, that makes it complicated for ArrayVec.. where a major version bump has larger ramifications..

bluss:
Thanks. I'll merge this & fix it up.

PR #27, opened 2016-02-06
Nodrop next

bluss:


bluss:
@gereeter nodrop 0.1.6 is now no-std-enabled. It required fixing the dependency odds first too.

Issue #32, opened 2016-03-02
Why does ArrayVec not derive Copy?


mitchmindtree:
I'm considering using an `ArrayVec<[_; 9]>` as a small stack to keep track of pressed buttons on a mouse in conrod, however I noticed that it does not impl `Copy`. Seeing as it has a fixed-size, shouldn't deriving `Copy` be ok? Or is there some other boundary stopping this?

bluss:
It's a relevant question. `ArrayVec` implements `Drop`, which precludes Copy.

I have considered making a Copy-only ArrayVec, but most of the hard things this library does are for the non-Copy part to be correct. The Copy case is simple, so a regular array is often enough.

WiSaGaN:
Would you accept a `CopyArrayVec` pull-request?

bluss:
I'm not sure, it requires duplicating the whole thing. As said above, the most interesting things ArrayVec does is handle the non-Copy case correctly.

I guess this is not what you want, but maybe specialization (unstable feature) can be used to make ArrayVec Copy exactly when its elements are Copy.

bluss:
a PoC of copy-specialization works. This is the ideal approach since it transparently removes the overhead of non-copy arrayvecs (overwrite in drop etc). The use of unstable features is rather non-ideal though.

mitchmindtree:
This issue isn't very high priority for me, I'm happy waiting until specialization stabilises if that suits you best :+1:

bluss:
We don't even know if specialization will allow this, but we'll see.

bluss:
It sounds like this is no longer just a question but a feature request?

tbu-:
I'm not sure that this needs a breaking change.

bluss:
arrayvec as a crate needs it. It's part of clearing the air before adding new features, if you will.

bluss:
Where the most significant breaking change is the change of minimum Rust version

c410-f3r:
What is the status of this feature?

smessmer:
Is this maybe easier to do now with the new array based implementation in 0.7?

pwnorbitals:
I’d be interested in this feature, is #193 on the right way to be merged ? thanks :)

PR #34, opened 2016-04-26
Add `ArrayVecCopy`, which is a copyable `ArrayVec`

tbu-:
Also add a common "backend" for `ArrayVec`, `ArrayString` and `ArrayVecCopy`.


bluss:
This looks interesting, good split. I indicated in #32 I would probably not accept this PR, but I'll need some time to look at it.

bluss:
I would like to revive this. The specialization trail doesn't seem to work out properly yet

tbu-:
So I can rebase this with a chance of approval?

bluss:
Yes.

Please revisit this change assert!(mem::size_of::<QuadArray>() <= 28);

in nightly we should have no drop flags and it should not grow at all. And I think that's a requirement, we don't want arrayvec to grow more.

src/copy.rs
```diff
@@ -0,0 +1,377 @@
...
+        ArrayVecCopy { inner: RawArrayVec::from_iter(iter) }
+    }
+}
+
+impl<A: Array + Copy> Clone for ArrayVecCopy<A>
+    where A::Item: Clone
+{
+    #[inline]
+    fn clone(&self) -> Self {
+        ArrayVecCopy { inner: self.inner.clone() }
```
bluss: this can be just `*self`

tbu-:
Rebased. Tightened the assertion.

src/copy.rs
tbu-: The current implementation can potentially copy less.

bluss: Hm, ok, so the rest of it is uninitialized. Great.

tbu-: Yea, but this is already the case for any new `ArrayVec`, all the elements are uninitialized there.

tests/tests.rs
```diff
@@ -130,15 +130,15 @@ fn test_is_send_sync() {
...
     type ByteArray = ArrayVec<[u8; 4]>;
     println!("{}", mem::size_of::<ByteArray>());
     assert!(mem::size_of::<ByteArray>() <= 8);
 
-    // 12 element size + 1 enum tag + 3 padding + 1 len + 1 drop flag + 2 padding
+    // 12 element size + 1 enum tag + 3 padding + 1 len + 3 padding
     type QuadArray = ArrayVec<[u32; 3]>;
     println!("{}", mem::size_of::<QuadArray>());
-    assert!(mem::size_of::<QuadArray>() <= 24);
+    assert!(mem::size_of::<QuadArray>() <= 20);
```
bluss: great that the new wrapper types don't cost any size anymore (no drop flags), but this will need to stay 24 to pass on stable as I guess you saw from travis

tbu-: On stable this will be 28 though.

bluss: I just realized. I am a bit slow today! That's not good.

bluss: Isn't the reason that NoDrop is now around both the array and the its length?

tbu-: Yes.

tbu-: We could declare that regression as fine, because it'll go away eventually (without further intervention).

bluss: I want to see if it can be solved

bluss:
Trying a second time in #40

tbu-:
It seems that drop flags are removed from stable, at least if I read the [release announcement][1] correctly. Would you be interested in a rebased patch, now that it doesn't regress on stable anymore?

[1]: https://blog.rust-lang.org/2016/11/10/Rust-1.13.html

bluss:
I'd like to get ArrayVecCopy in, but I still prefer the trait based approach. But.. now I'm just trying to find out which of the two approaches that squares better with [this kind of representation for arrayvec](https://internals.rust-lang.org/t/mem-uninitialized-and-trap-representations/4167/48)

tbu-:
I would redo this if desired. I'm still looking for an `ArrayVecCopy` for my project.

tbu-:
The regression is long gone.

Meh. Can you tell me whether I can close this pull request? There's still need for a copyable array vec in the ecosystem (see e.g. https://www.reddit.com/r/rust/comments/5zvve7/arrayinit_safe_array_initialization/), but this doesn't seem to move, neither does your second PR in #40.

bluss:
Right, I haven't had time. If I use #40, can I use your commits there? To be practical about myself, that's the one that be likely to merge.

It's not good that this stops with me, I need to figure out how to bring other people on board to take care of the project.

tbu-:
Yes, you can use my commits there if you want. (Not sure if they're helpful though, probably easier to rewrite.)

I'd also redo this pull request if you would merge it.

bluss:
Oh right, so to get this on track. There are unreleased changes on the master branch

- arrayvec will have another 0.3.x release
- merge arrayvec copy, bump the minimum Rust version and release a new major version (0.4.x)

bluss:
Any wishes in particular when it comes to minimum supported Rust version?

bluss:
Oh I had forgotten that we have generic_array as public dependency too. Need to check up on that.

tbu-:
Redoing this PR now. I don't think we need a new minimum Rust version, the older ones will continue to work, although they may use a little more memory. Minimum version for the current amount of memory seems to be 1.13.0 according to my comment in this pull request.

bluss:
I didn't understand the amount of memory thing. Current minimum version for arrayvec is Rust 1.2 according to travis, that's the one I want to change.

tbu-:
Rust versions prior to 1.13 had a drop flag inserted in every struct implementing `Drop`. This is no longer the case, which is why this pull request is no longer worse on memory than the current code.

Users with a Rust version smaller than 1.13 will have a regression of memory usage (not much, it's one more drop flag for the `ArrayVec` struct).

bluss:
I don't think you should redo this PR. This is based on that I said I prefer #40. I don't want you to feel like you are wasting time (again).

tbu-:
Ah, thanks for clarifying. I somehow misunderstood you there.

PR #39, opened 2016-09-21
Fix bounds checking in ArrayVec::insert(index, element)

### LINKED ISSUE #38: `insert` does no bounds checking
The function looks wrong for cases `self.len() < index < self.capacity()`.

bluss:
Must be `index <= len && index < capacity`

Fixes #38

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -179,6 +179,10 @@ impl<A: Array> ArrayVec<A> {
     ///
     /// Return `None` if no element is shifted out.
     ///
+    /// `index` must be <= `self.len()` and < `self.capacity()`. Note that any
+    /// out of bounds index insert results in the element being "shifted out"
+    /// and returned directly.
+    ///
     /// ```
     /// use arrayvec::ArrayVec;
     ///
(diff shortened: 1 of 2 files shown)


bluss:
Oh, there's more broken than that. Let's fix overwriting too.

tbu-:
I think panicking would be more appropiate, according to the documentation and my gut feeling. No element is pushed out if we're passing an out-of-range index.

Otherwise the documentation should probably be updated to reflect that the inserted element is returned for out-of-bounds inserts.

bluss:
You have a good point.

bluss:
I'm not comfortable with doing that, it will simply be something to consider for the next major version. This version has the weird shift out behavior already (and the tests ensure it).

bluss:
Current tests try things like:

```
     let mut v = ArrayVec::from([2]);
     assert_eq!(v.insert(1, 1), Some(1));
     assert_eq!(v.insert(2, 1), Some(1));
```

tbu-:
Right.

PR #40, opened 2016-09-21
Arrayvec copy 2

bluss:
WIP continuation of #34

Using traits instead of data for reusing code


bluss:
@tbu- What do you think? This solves it without size expansion. I think I like the trait for code reuse better.

tbu-:
This works too.

I view it as a complication of just sharing the same struct, if the no-regressions thing is that important, we could also wait a cycle (or two?) until the no-drop-flags thing hits stable.

PR #41, opened 2016-09-26
Support generic-array as arrayvec backend

bluss:
I'm sorry, I never realized we could do this as an addition, without changing how arrayvec works.

This is awesome!

This adds generic-array as an optional dependency. When it is, you can use GenericArray as an arrayvec storage backing.


bluss:
cc @fizyk20 this is cool

PR #42, opened 2016-10-29
Array deque

goandylok:
I was looking for a stack-only ring buffer, but none of them is sufficient for my work. 

I noticed this crate before, as it provided a nice pattern to create stack-only collections, at the meanwhile, I don't want to duplicate the Array trait implementation, so I tried making this pull request.

However, some refactor on module relationship may needed.

### DIFF
--- a/src/array_deque.rs
+++ b/src/array_deque.rs
@@ -656,7 +656,7 @@ impl<A: Array> ArrayDeque<A> {
     /// assert_eq!(vector.contains(&10), false);
     /// ```
     pub fn contains(&self, x: &A::Item) -> bool
-        where A::Item: PartialEq<A::Item>
+        where A::Item: PartialEq
     {
         let (a, b) = self.as_slices();
         a.contains(x) || b.contains(x)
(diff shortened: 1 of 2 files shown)


goandylok:
It seems that rustc 1.2.0 cannot handle associated type perfectly.

goandylok:
It seems that rustc 1.2.0 cannot handle associated type perfectly.

bluss:
Don't bother with Rust1.2 IMO.

This looks well written, it is wise to reuse parts of rusts own implementation. Can you please consider publishing this as its own crate?

With untagged unions, we are approaching a complete rewrite of the core parts of arrayvec, I don't want to expand our codebase so much ahead of that. Thanks.

bluss:
Don't bother with Rust1.2 IMO.

This looks well written, it is wise to reuse parts of rusts own implementation. Can you please consider publishing this as its own crate?

With untagged unions, we are approaching a complete rewrite of the core parts of arrayvec, I don't want to expand our codebase so much ahead of that. Thanks.

goandylok:
Never mind. I've published the crate [`arraydeque`](https://github.com/goandylok/arraydeque).

PR #52, opened 2017-04-22
Add `PartialOrd` and `Ord` implementation to `ArrayString`

### LINKED ISSUE #50: Could `ArrayVec` and `ArrayString` implement `Ord`?
I would love to use `ArrayString`, but it doesn't implement `Ord` which is a constraint for me (types in differential dataflow need to be `Ord` so they can be sorted for deduplication/consolidation). 

It seems like it could be a pretty simple implementation, as `String` just derives `Ord` which means it gets the `Vec<u8>` implementation. I guess it would be similar here, except needing to manually slice the backing contents in the impl to avoid touching invalid data. It seems like `ArrayVec` could be similarly workable, for types `T: Ord`.

Edit: I'd be happy to do a PR, but I'm not sure I grok all of the index cleverness, and wouldn't want to mess things up. If you'd rather I take a stab first, let me know. :)

tbu-:
Fixes #50.

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -1,4 +1,5 @@
 use std::borrow::Borrow;
+use std::cmp;
 use std::fmt;
 use std::hash::{Hash, Hasher};
 use std::mem;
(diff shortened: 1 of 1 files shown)


tbu-:
Requested in #51.

bluss:
Thanks!

PR #53, opened 2017-05-13
Factor out push_unchecked()

niklasf:
This can be useful when it's known in advance that the capacity is sufficient.


bluss:
Thanks, this is nice. I'd use a debug assertion in push_unchecked, which helps our tests as well as the end user during development.

niklasf:
Makes sense. Added.

bluss:
Thanks. Sorry for the delay.

PR #55, opened 2017-06-08
Add serde support

### LINKED ISSUE #54: Serde support
It would be nice if `ArrayVec`/`ArrayString` had their optional `serde` `Serialize` and `Deserialize` implementations.

daboross:
This implements serde support under the optional 'serde' feature, and adds unit tests to test said support.

Serde will be optionally included for runtime, but is unfortunately required for running tests - as cargo does not have support for optional dev-dependencies. `FEATURES="serde"` is added to the travis config, as serde-support tests will not run without it.

Fixes #54.

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -14,6 +14,9 @@ use array::Index;
 use CapacityError;
 use odds::char::encode_utf8;
 
+#[cfg(feature="serde")]
+use serde::{Serialize, Deserialize, Serializer, Deserializer};
+
 /// A string with a fixed capacity.
 ///
 /// The `ArrayString` is a string backed by a fixed size array. It keeps track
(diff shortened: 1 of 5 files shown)


daboross:
Removed usage of `?`.

daboross:
Sorry about that - removed usages of '?' in the code.

daboross:
@bluss Should I re-make this PR towards the `0.3` branch instead?

bluss:
Hm probably not. We should aim forwards.

I'm thinking here I'd prefer if the feature flag was versioned and had a separate name from the crate. So a feature name like `serde-1` or `use_serde` would sit better. I guess the standard is just "serde" but I've seen crates where that is a problem.

daboross:
@bluss Ah, just missed the comment!

I was basing the `serde` name off of https://github.com/brson/rust-api-guidelines/blob/master/README.md#data-structures-implement-serdes-serialize-deserialize-c-serde, which states that the feature should be the crate name itself.

bluss:
Thanks for the link, I don't think it's that simple, even though it's a neat solution for that particular issue. It's time to break the feature and optional crate namespace merge. Using the crate name limits options when it comes to the evolution of the crate and avoiding breaking changes.

bluss:
I for example think that ndarray 0.9, which supports serde 0.9, it could have added support for serde 1.0 without breaking change if it had the foresight.

daboross:
Ah, that makes sense. I wasn't sure if depending on multiple versions of a crate would be possible even under different optional features. `serde-1` is good then?

bluss:
Right -- it's not straightforward to do such a dependency. But the start of it exists :smile: [as legacy-serde](https://crates.io/crates/legacy-serde). Maybe the idea is not fully matured.

I think serde-1 is the way to go.

daboross:
Sounds good then! Feature name updated.

dtolnay:
If you are versioning the feature to possibly support multiple Serde versions in the same arrayvec in the future, I would prefer to standardize on the official shims. Add a dependency on [`serde1`](https://crates.io/crates/serde1) and implement `serde1::{Serialize, Deserialize}`. That way no implicit `"serde"` feature is inferred by Cargo.

src/array_string.rs
```diff
@@ -324,3 +327,48 @@ impl<A: Array<Item=u8>> Ord for ArrayString<A> {
...
+        use serde::de::{self, Visitor};
+        use std::marker::PhantomData;
+
+        struct ArrayStringVisitor<A: Array<Item=u8>>(PhantomData<A>);
+
+        impl<'de, A: Array<Item=u8>> Visitor<'de> for ArrayStringVisitor<A> {
+            type Value = ArrayString<A>;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                write!(formatter, "a string with no more than {} elements", A::capacity())
```
dtolnay: The word "element" of a string is unclear to me. Is this terminology used elsewhere in arrayvec? I would prefer to say "a string of no more than {} *bytes*".

daboross: Sure, that makes more sense.

daboross:
Thank you for the comment about the messages @dtolnay, I think the string message was really just copy-pasted from the ArrayVec one. Should be more clear now!

daboross:
@bluss thoughts on depending on `serde1` vs. just `serde` with `serde-1` feature?

daboross:
When implementing a failing test (idea from #58), I found what seem to be a bug in either this code or in serde's error code?

When deserializing an ArrayVec that's too large, the message is
```
invalid type: integer `13`, expected an array with no more than 2 items
```

where I'd expect it to be
```
invalid length: 3, expected an array with no more than 2 items
```

Code producing this error:

```rust
fn visit_seq<SA>(self, mut seq: SA) -> Result<Self::Value, SA::Error>
    where SA: SeqAccess<'de>,
{
    let mut values = ArrayVec::<A>::new();

    while let Some(value) = try!(seq.next_element()) {
        if let Some(_) = values.push(value) {
            return Err(SA::Error::invalid_length(A::capacity() + 1, &self));
        }
    }

    Ok(values)
}
```

daboross:
Just added failure tests, thanks to #58 for the idea.

P.S. Sorry, ignore my last comment - I just had a bug in the test code I wrote.

alexbool:
@bluss friendly ping

bluss:
Well, I'm certainly being unfriendly by not being on top of this. (Not guaranteed to be here at any one particular month of the year, but some of them would be nice...) Let's see..

bluss:
Comaintainers / maintainers that are interested in taking care of this crate? I'm not planning on stepping down, but I clearly don't have enough time to do the whole job.

bluss:
Many thanks for the PR

bluss:
Features are in a deadlock in arrayvec, since the 0.3 branch says the version requirement is Rust 1.2.0. Ridiculous, sure. So we can't release any new features on that branch. (Master already tests from newer versions in travis.)

Hoping to resolve all the planned API changes for the next version and release 0.4. It's not ideal, but the issues are pretty long standing as well.

PR #57, opened 2017-06-08
Add serde support to 0.3

### LINKED ISSUE #54: Serde support

daboross:
This implements serde support under the optional 'serde' feature, and adds unit tests to test said support.

Serde will be optionally included for runtime, but is unfortunately required for running tests - as cargo does not have support for optional dev-dependencies. FEATURES="serde" is added to the travis config, as serde-support tests will not run without it.

See #55 for master PR.

Fixes #54.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ matrix:
     - rust: 1.2.0
     - rust: stable
       env:
-        - FEATURES="use_generic_array"
+        - FEATURES="use_generic_array serde"
         - NODEFAULT=1
     - rust: beta
     - rust: nightly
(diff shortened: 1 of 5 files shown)


daboross:
Doesn't seem like this is possible to add while supporting 1.2.0, even as an optional dependency.

bluss:
Yeah that's quite the quirk; serde broke old cargo unconditionally :-)   Just shows that so old Rust deps are unsustainable. Let's hope the next 15 Rust versions fare better w.r.t backwards compat.

bluss:
Which is part of the reason why everything arrayvec is stalling

- Can't put in new features in version 0.3.x due to Rust version requirement
- Lots of proposed breaking changes that need to get in before the next major version

PR #58, opened 2017-06-11
Add serde support.

camlorn:
Add support for serde to both ArrayVec and ArrayString.

I think I did this right.  The 6 tests I added all pass.  The downside is that we have to make serde  a default feature in order to get tests working, in so far as everyone on #rust-offtopic knew.  This isn't a major problem since Serde doesn't force us to use std.

My motivating use case is an implementation of a disk-backed B+ tree.  This also makes it possible to use Arrayvec  to avoid allocations with networking applications when using Serde to encode and decode packets.

There might be formatting problems.  If there are, let me know and I'll fix them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,16 @@ version = "0.1.8"
 path = "nodrop"
 default-features = false
 
+[dependencies.serde]
+version = "1.0"
+default-features = false
+optional = true
+
 [features]
-default = ["std"]
-std = ["odds/std", "nodrop/std"]
+default = ["std", "use_serde"]
+std = ["odds/std", "nodrop/std", "serde/std"]
 use_union = ["nodrop/use_union"]
+use_serde = ["serde"]
+
+[dev-dependencies]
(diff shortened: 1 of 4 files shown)


camlorn:
Looks like this is duplicate work to and possibly an inferior version of #55.

I'm leaving this open for now, but we'll probably end up closing it.

daboross:
Doesn't look too bad! I think the separate file for the implementations might be a good idea. Having tests to test failure is also something I missed - would it be alright with you if I added that idea to #57?

For the tests, my PR just has the whole test file with `#![cfg(feature = "serde-1")]` so that the tests won't run if `--features serde` isn't enabled. I imagine this could be done here too, if you want.

camlorn:
Absolutely.  I'm not attached to this PR.  I just needed it for one of my own projects and implemented it without checking beyond seeing that it wasn't there.

(also, I thought it would take half an hour. It didn't).

bluss:
Superseded by #55, thanks anyway!

PR #59, opened 2017-06-15
Add Array implementation for array sizes 50, 100 and 200

daboross:
The implementation for 50 would be useful for a use case of mine, if that's alright. 100 and 200 are added for completeness to match 50 existing.


bluss:
Yes, why not. We'll roll this in. Thank you and sorry for the delay.

daboross:
Thank you, and no problem! Glad this is alright to add.

PR #61, opened 2017-07-30
Prepare 0.4, more vec-like api and error reform

### LINKED ISSUE #45: Consider panicking in `ArrayVec::remove` instead of returning an `Option`

### LINKED ISSUE #46: Consider returning an error instead of an option when full

bluss:
## `push, insert` etc

With the goal of making arrayvec more of a "drop-in" for Vec, we realign the signatures methods `.push(elt), .insert(index, elt)` and `.remove(index)`.

```rust
pub fn push(&mut self, element: A::Item);
pub fn insert(&mut self, index: usize, element: A::Item);
pub fn remove(&mut self, index: usize) -> A::Item;
pub fn swap_remove(&mut self, index: usize) -> A::Item;
```

## `try_push` etc

Mimicing something like fallible allocation, we have the fallible new methods:

```rust
pub fn try_push(&mut self, element: A::Item) -> Result<(), CapacityError<A::Item>>;
pub fn try_insert(&mut self, index: usize, element: A::Item) -> Result<(), CapacityError<A::Item>>;
```

Notice that `try_insert` still panics if `index` is out of bounds.

## `pop_at` etc

Mimicing `.pop()`, we have new “checked” versions of `remove, swap_remove`:

```rust
pub fn pop_at(&mut self, index: usize) -> Option<A::Item>;
pub fn swap_pop(&mut self, index: usize) -> Option<A::Item>;
```

The rationale is: indexing out of bounds is not a regular runtime error, it's normally a contract violation. So when we introduce checked versions of such API, then we get “*checked*” versions that return `Option`. Non-presence is not necessarily an error.

## `ArrayString::push` etc

- ArrayString: push, push_str now have the same signature as corresponding on String. New methods try_push, try_push_str.

```rust
pub fn push(&mut self, c: char);
pub fn try_push(&mut self, c: char) -> Result<(), CapacityError<char>>;
pub fn push_str(&mut self, s: &str);
pub fn try_push_str<'a>(&mut self, s: &'a str) -> Result<(), CapacityError<&'a str>>;
```

Closes #46 
Closes #45

## Miscellaneous Changes

- odds is no longer a public dependency. We want the flexibility of dropping/upgrading this dep at will
- Rust version changes

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,7 +13,9 @@
 //!   - Use the unstable feature untagged unions for the internal implementation,
 //!     which has reduced space overhead
 //!
-//! **arrayvec Requires Rust 1.18**
+//! ## Rust Version
+//!
+//! This version of arrayvec requires Rust 1.15 or later.
 //!
 #![doc(html_root_url="https://docs.rs/arrayvec/0.3/")]
 #![cfg_attr(not(feature="std"), no_std)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -95,6 +95,13 @@ impl<A: Array> Drop for ArrayVec<A> {
     }
 }
 
+macro_rules! panic_oob {
+    ($method_name:expr, $index:expr, $len:expr) => {
+        panic!(concat!("ArrayVec::", $method_name, ": index {} is out of bounds in vector of length {}"),
+               $index, $len)
+    }
+}
+
 impl<A: Array> ArrayVec<A> {
     /// Create a new empty `ArrayVec`.
     ///
@@ -278,7 +285,7 @@ impl<A: Array> ArrayVec<A> {
     /// ```
     pub fn try_insert(&mut self, index: usize, element: A::Item) -> Result<(), CapacityError<A::Item>> {
         if index > self.len() {
-            panic!("ArrayVec::try_insert: {} is out of bounds (length is {})", index, self.len());
+            panic_oob!("try_insert", index, self.len())
         }
         if self.len() == self.capacity() {
             return Err(CapacityError::new(element));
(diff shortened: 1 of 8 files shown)


tbu-:
I'd prefer methods that panic on invalid indices and return an error on capacity error. Maybe we could also ask other developers about this. I think the proposed fallible allocation API on internals.rust-lang.org also behaves that way.

bluss:
Thanks for the feedback. Even if this crate is moving at glacial pace.

From my https://github.com/bluss/arrayvec/pull/55#issuecomment-318891667

> Comaintainers / maintainers that are interested in taking care of this crate? I'm not planning on stepping down, but I clearly don't have enough time to do the whole job.

tbu-:
I'd be interested in taking care of this crate. What do you think?

bluss:
Sure. We can do it together? I'm not sure what you have in mind. In one way we want to guide this crate towards 1.0, but we're missing a few pieces from Rust for that (ManuallyDrop in Rust 1.20 and uninit data in inline field clarification is ??). With API overhaul it feels like we are ready with every we can do on our side.

(Clearly integer generics is not something to wait for 1.0 for)

bluss:
Sure, that sounds good!

I had some discussion on #rust-libs and they sort of brought me over to mixing panic and result for insert.

If we think of arrayvec as a Vec with a particular inline allocator, we have the problem of fallible allocation (capacity is limited); and a hypothetical fallible allocation `Vec::try_insert` would panic on out of bounds, but return an error for allocator errors.

It's not obvious that fallbility of insert is just about allocation/capacity, but with the comparison to Vec it seems ideal to make it that way.

And thus another reminder that arrayvec is obsolete once custom allocators get so far that Vec can be backed by an array! That sounds cool.

bluss:
Sigh, I don't know anymore if it's best to have the error cases in push or in the separate fallible version. Both those seem fine

tbu-:
I'd prefer the non-prefixed ones, but that's no strong opinion.

Do you want to finish this API change yourself or should I do it?

bluss:
I can finish it

bluss:
ping @tbu- if you want. This is getting better with the feedback IMO. Most of your thoughts have been incorporated. PR description updated.

src/lib.rs
```diff
@@ -285,42 +339,94 @@ impl<A: Array> ArrayVec<A> {
...
         self.pop()
     }
 
     /// Remove the element at `index` and shift down the following elements.
     ///
-    /// Return `Some(` *element* `)` if the index is in bounds, else `None`.
+    /// The `index` must be strictly less than the length of the vector.
+    ///
+    /// ***Panics*** if the `index` is out of bounds.
+    /// vector.
```
tbu-: Extra word.

src/lib.rs
```diff
@@ -95,6 +98,13 @@ impl<A: Array> Drop for ArrayVec<A> {
     }
 }
 
+macro_rules! panic_oob {
+    ($method_name:expr, $index:expr, $len:expr) => {
+        panic!(concat!("ArrayVec::", $method_name, ": index {} is out of bounds in vector of length {}"),
+               $index, $len)
+    }
+}
```
bluss: Right. The approach is to more or less go for correct with this PR and safe fast for later. I'm not sure we can find a difference, but it would be nice to at least put the panic formatting code in a place that is independent of the type parameters.

src/lib.rs
bluss: Thanks

bluss:
Yes -- the unwraps have room for improvement. They should all be Result::unwrap's so they have their error information there in a basic form.

bluss:
Let's go, so that we finally clear this hurdle.

bluss:
New rust version requirement is open to suggestions... Right now I can't imagine that the current era Rust versions are going to break like Rust 1.0 - 1.10 ish seem to have done, but maybe I'll be surprised.

PR #63, opened 2017-07-31
Add a faster retain that does not preserve order

niklasf:
Feature idea: Without preserving the order elements can be filtered much faster.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -367,6 +367,31 @@ impl<A: Array> ArrayVec<A> {
         }
     }
 
+    /// Retains only the elements specified by the predicate.
+    ///
+    /// It is faster than `array.retain()` but does not preserve the order
+    /// of the retained elements.
+    ///
+    /// ```
+    /// use arrayvec::ArrayVec;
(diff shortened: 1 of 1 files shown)


tbu-:
This is only faster for cases where the majority of the elements stay in the array. For the other case, the other function should be faster.

niklasf:
Even if just a single element at index `i` is removed, `retain()` has to move `len() - 1 - i` elements up one position. `swap_retain()` will just move the element from the end into that position.

tbu-:
In the source code of the `retain` function, you can see that it does one swap for every element that is retained (after at least one element isn't). This means that for the case where all elements are removed (the best case for `retain`), it does exactly zero swaps. This variant does `n` swaps instead. Of course, if more elements are retained than are removed, this new function is likely faster.

niklasf:
Thanks for the explanation. That's right of course. Feel free to close if the remaining use case is too specialized.

bluss:
I think we can just leave this out; it's simple to implement outside.

PR #65, opened 2017-08-06
Prepare 0.4

bluss:
I think everything is finally ready to go.

It's hard to pick any particular minimal Rust version (1.15 picked now). We don't want to ignore everyone who is a few versions behind, but we also don't want to get into the hole that was requiring Rust 1.2 ever again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ env:
   - FEATURES='serde-1'
 matrix:
   include:
-    - rust: 1.15.0
+    - rust: 1.14.0
     - rust: stable
       env:
         - NODEFAULT=1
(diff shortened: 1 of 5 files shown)


heinrich5991:
I suggest Rust 1.14.0, it is the [Rust version in Debian stable](https://packages.debian.org/stretch/rustc).

bluss:
Ok, let's see if we can jump to Rust 1.14. ManuallyDrop is not really ready yet (stable & sound with our use) but if it is, we might jump to Rust 1.20 later.

bluss:
It passed, stopped the redundant build.

tbu-:
🎉

Issue #66, opened 2017-08-29
ArrayString lacks removal features


james-darkfox:
`ArrayString` lacks any removal features. `pop` `truncate` `remove` etc would be nice.

All of these may be copied from `std::String` AFAIK.

I will implement these and more if requested. I've opened this issue to discuss because this seems too obvious to have been accidentally omitted?

tbu-:
These are not omitted on purpose.

james-darkfox:
@tbu- I would expect not. I'll implement this "soon (tm)" unless someone beats me to it.

bluss:
Mentoring instructions: PRs are welcome.

- Focus on mimicking existing `String` methods in their complete API (equal signature is preferred) and in docs (adapted appropriately).
- Ask me on rust channels on mozilla IRC if you want

PR #68, opened 2017-09-11
Add `Default` implementation for `ArrayString`

### LINKED ISSUE #67: Should `ArrayString` implement `Default`?
I don't know that it should, but `ArrayVec` does, and `String` does. I just noticed that `ArrayString` does not, and while it isn't blocking for me (in fact it reminded me that I should remove the `: Default` requirements on some of my impls) I thought I should point it out!

tbu-:
Fixes #67.

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -31,6 +31,12 @@ pub struct ArrayString<A: Array<Item=u8>> {
     len: A::Index,
 }
 
+impl<A: Array<Item=u8>> Default for ArrayString<A> {
+    fn default() -> ArrayString<A> {
+        ArrayString::new()
+    }
+}
+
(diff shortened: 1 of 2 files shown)


tbu-:
@bluss Can I merge small changes like this myself, or would you rather review them?

bluss:
Let's start out by reviewing everything I think. Maybe I'm a bad sport, because I just pushed three doc changes just before I saw this, sorry about that. Nothing is really critical until we make releases anyway. I'll approve this and you can merge though..

bluss:
I can release 0.4.1 with this and the doc fixes. Anything else we want to get in?

tbu-:
Should be good.

We want #66 at some point, but I don't have an implementation yet.

bluss:
Great, yes those should be simple to add.

PR #71, opened 2017-09-26
Make zero-capacity ArrayVec a zero-sized type.

clarcharr:
This changes the `Index` for zero-length `Array`s to be `()`, ensuring that they are zero-sized. Additionally, this adds `bool` as the index for singleton `Array`s, to become compatible with future enum optimisations, which could potentially make `Option<ArrayVec<[u8; 1]>>` be 16 bits long instead of 32 once more vigorous enum discriminant optimisations are implemented.

### DIFF
--- a/src/array.rs
+++ b/src/array.rs
@@ -32,6 +32,20 @@ pub trait ArrayExt : Array {
 
 impl<A> ArrayExt for A where A: Array { }
 
+impl Index for () {
+    #[inline(always)]
+    fn to_usize(self) -> usize { 0 }
+    #[inline(always)]
+    fn from(ix: usize) ->  Self { () }
+}
+
+impl Index for bool {
+    #[inline(always)]
+    fn to_usize(self) -> usize { self as usize }
+    #[inline(always)]
+    fn from(ix: usize) ->  Self { ix != 0 }
+}
+
 impl Index for u8 {
     #[inline(always)]
     fn to_usize(self) -> usize { self as usize }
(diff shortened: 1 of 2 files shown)


tbu-:
Looks good to me. Since this is a breaking change, I suggest to only merge this with a new version.

bluss:
Fun trick -- do you have a use case?

clarcharr:
I currently don't, although it was something I noticed and I figured I'd give it a try. Doesn't seem like too much of a maintenance burden and I figured it'd be useful to someone.

bluss:
Don't worry. It's not a burden to carry, but breaking changes and release planning are a burden, so it would be more important if it did have a use case. I wouldn't want to merge this without release planning, since from experience, breaking changes in master are in the way if we need or want to do compatible fixes. That said, thanks a lot for coming up with this idea before we we reached 1.0.

clarcharr:
Yeah, definitely! I wasn't expecting this to get merged immediately; I was just offering the change because I think it's a good one. :)

clarcharr:
@bluss What's blocking this being merged right now?

bluss:
It's a breaking change, so it can be merged when we are ready for the next release. I don't think we are waiting on anything else though, so it could be anytime. I don't have time today.

clarcharr:
When I submitted #104 I also rebased this branch so it would be compatible with the latest changes. Not sure if there's any desire to merge this at some point, but I figured I'd mention it.

bluss:
We'll merge it, I just haven't been in Rust a lot this year. Right now I have time, and the time has come to arrayvec.. :slightly_smiling_face:

bluss:
Fixing the conflict I caused. Convenient, but I hope you think that's ok..

bluss:
Thanks!

bluss:
breaking change season. And fixing the manuallydrop stuff

PR #75, opened 2017-10-18
Add ArrayVec::truncate()

niklasf:
This would be nice to have for symmetry with [Vec::truncate()](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.truncate).


bluss:
cc #66 

Thank you, this is good, follows Vec!

bluss:
Oh 66 was about array string, not array vec. Oh well, we can ping that issue for fun anyway

PR #76, opened 2017-10-20
Prototype for MaybeUninit formulation of arrayvec

bluss:
- Testing ideas for arrayvec 1.0
- Looking at what is well defined and what's not
- Not adequate yet:
  - How to handle uninitialization?
  - How to handle ArrayString? It's Copy & uninitialized (always an array of u8)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,34 +4,27 @@ env:
   - FEATURES='serde-1'
 matrix:
   include:
-    - rust: 1.14.0
+    - rust: 1.21.0
     - rust: stable
       env:
         - NODEFAULT=1
-        - NODROP_FEATURES='use_needs_drop'
     - rust: beta
     - rust: nightly
       env:
         - NODEFAULT=1
     - rust: nightly
-      env:
-      - NODROP_FEATURES='use_needs_drop'
     - rust: nightly
       env:
-      - FEATURES='serde use_union'
-      - NODROP_FEATURES='use_union'
+      - FEATURES='serde'
 branches:
   only:
     - master
     - 0.3
 script:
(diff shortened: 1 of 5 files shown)


bluss:
cc @arielb1 and thanks a lot for the help in searching for the best expression of this in Rust

tbu-:
Is the `repr(C)` supposed to inhibit the enum optimization?

bluss:
`repr(C)` was put there only there so that we can cast the `*const MaybeUninit<T>` to `*const T`.  There is a missing link right now, that needs to be rationalized as sound for the intermediate `ManuallyDrop` too.

bluss:
A question I didn't fully understand if it was answered:

If we have a `MaybeUninit<[T; 3]>` which is partially initialized; only its first element is initialized.

1. Is it sound to extract a `&[T; 3]` from this?
2. Must arrayvec instead extract a `*const T`?
  * With the eventual goal of course, of creating `&T` to an element only if that element is initialized.

bluss:
Yesterday's discussion https://botbot.me/mozilla/rust-libs/2017-10-20/?msg=92551510&page=1

tbu-:
Thanks for the link to the discussion. Sounds like having a `&[T; 3]` with uninitialized elements is UB or at least not clearly not UB. We could work with `*mut [T; 3]` and `*mut [T]`.

bluss:
What we were trying to do is to get docs amended for ManuallyDrop so that it can be MaybeUnitialized, but that seems to not be such a good idea. I'm still not entirely convinced.

bluss:
New version in #97

PR #79, opened 2017-10-26
Promise debug assertions for unsafe methods

bluss:
Instead of being vague about it, we can promise it.

ArrayString::set_len is a bit more complicated. I've already removed the char boundary check since the first revision of the PR.

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -322,10 +322,11 @@ impl<A: Array<Item=u8>> ArrayString<A> {
 
     /// Set the strings's length.
     ///
-    /// May panic if `length` is greater than the capacity.
-    ///
     /// This function is `unsafe` because it changes the notion of the
     /// number of “valid” bytes in the string. Use with care.
+    ///
+    /// This method uses *debug assertions* to check the validity of `length`
+    /// and may use other debug assertions.
     #[inline]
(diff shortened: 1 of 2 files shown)


bluss:
r? @tbu- 

The new debug check in ArrayString is potentially breaking, so I'm not 100% sure if it should go into a 0.4.x update, what do you think?

src/array_string.rs
```diff
@@ -322,14 +322,16 @@ impl<A: Array<Item=u8>> ArrayString<A> {
...
     /// This function is `unsafe` because it changes the notion of the
     /// number of “valid” bytes in the string. Use with care.
+    ///
+    /// This method uses a *debug assertion* to check the validity of
+    /// `length` (character boundary and inside capacity).
     #[inline]
     pub unsafe fn set_len(&mut self, length: usize) {
         debug_assert!(length <= self.capacity());
         self.len = Index::from(length);
+        debug_assert!(self.is_char_boundary(length));
```
bluss: Oh sigh this is actually broken. `self as &str` depends on length, and the current length is always a character boundary...

tbu-:
With just the added guarantees in the documentation, this looks uncontroversial to me. Looks good.

Issue #73, opened 2017-10-03
Remove dependency on odds crate


clarcharr:
The only reference to this crate is via the `encode_utf8` function, which is now part of the standard library. IMHO it'd be better to use the stdlib version instead of that one.

bluss:
Makes sense when it's removed from nodrop too.

bluss:
Man, that odds thing really needs a good spring cleaning.  In nodrop we only use its debug_assert_unreachable, and the annoying thing is that `std::mem::unreachable` (whatever its final name is going to be) is still unstable, and the crate unreachable depends on another micro crate void. Which is much smaller than odds, but still annoying :-)

clarcharr:
IMHO a lot of the things in odds should be separated into other crates or just sent upstream. It's kind of ridiculous.

bluss:
Yes I agree of course. It's partly happening in its master branch, but there's a bit much to clean up.

bluss:
Odds is showing signs of Rust gaining features, it contains obsolete versions of Vec::splice, std::ptr::eq, str::repeat and some more things

bluss:
arrayvec 0.4.3 depends on odds *without* std, which might save thenths of seconds of compile time.

bluss:
Nevermind nodrop puts that std back 🤦‍♂️

PR #80, opened 2017-10-26
nodrop: Remove odds dependency

bluss:
Copy in some code in an effort to avoid odds dependency in nodrop.

We could depend on crate unreachable, but it in turn depends on crate void, and that's a bit excessive.

cc #73


bluss:
This seems easy.

PR #81, opened 2017-10-27
Remove odds dependency in arrayvec

### LINKED ISSUE #73: Remove dependency on odds crate

bluss:
Copy the encode_utf8 function from odds. std encode_utf8 requires Rust
1.15 and has a different signature, this one seems to fit us better.

Fixes #73

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,10 @@ bencher = "0.1.4"
 name = "extend"
 harness = false
 
+[[bench]]
+name = "arraystring"
+harness = false
+
 [features]
 default = ["std"]
 std = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,10 +11,6 @@ repository = "https://github.com/bluss/arrayvec"
 keywords = ["stack", "vector", "array", "data-structure", "no_std"]
 categories = ["data-structures", "no-std"]
 
-[dependencies.odds]
-version = "0.2.23"
-default-features = false
-
 [dependencies.nodrop]
 version = "0.1.8"
 path = "nodrop"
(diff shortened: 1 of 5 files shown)


bluss:
cc @tbu- 

I can't decide what's better. The Result-return formulation seems better for ArrayString. Maybe it's not something we should care about, could use .encode_utf8() in Rust **1.15**

tbu-:
I think our version fits `ArrayString` better. As I'd also like to keep Rust 1.14.0 compatibility for Debian (unless something very useful gets stabilized), I think we should just use this one.

bluss:
Benchmarks for fun. I managed to get the char::encode_utf8 to have matching performance, so we don't need to decide on that.

```
test push_alpha      ... bench:           9 ns/iter (+/- 0) = 56888 MB/s
test push_c          ... bench:           0 ns/iter (+/- 0) = 512000 MB/s
test push_string     ... bench:       1,772 ns/iter (+/- 68) = 288 MB/s
test try_push_alpha  ... bench:          41 ns/iter (+/- 3) = 12487 MB/s
test try_push_c      ... bench:         110 ns/iter (+/- 8) = 4654 MB/s
test try_push_string ... bench:       1,717 ns/iter (+/- 70) = 298 MB/s
```

I'm not super concerned that some of the benchmarks "optimize out" here. That's a pretty good mark in a way, it shows the compiler can see through the implementation and optimize it well.

tbu-:
:+1:

bluss:
Cool. We can always consolidate later. This way we can ship it in the 0.4.x series (Which we hope is the last before 1.0, I think?)

PR #82, opened 2017-11-01
Add `ArrayVec::resize` and friends

### LINKED ISSUE #72: Missing ArrayVec::resize
I'm missing the resize functionality (similar to the std::vec::Vec::resize):

fn resize(&mut self, new_len: usize, value: T)

I have to fill some items in a random order, so first the array have to be resized with some default values before indexing can start.
I could use plain arrays, but I'd like to use ArrayVec for it's "virtual dynamic" size.

Is there a function for this I've missed?

tbu-:
`ArrayVec::resize_default`
`ArrayVec::resize`
`ArrayVec::try_resize_default`
`ArrayVec::try_resize`

Fixes #72.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,6 +35,7 @@ extern crate core as std;
 
 use std::cmp;
 use std::iter;
+use std::marker::PhantomData;
 use std::mem;
 use std::ptr;
 use std::ops::{
@@ -595,24 +596,23 @@ impl<A: Array> ArrayVec<A> {
     pub fn as_mut_slice(&mut self) -> &mut [A::Item] {
         self
     }
-}
 
-impl<A: Array> ArrayVec<A>
-    where A::Item: Clone,
-{
-    /// Resizes the `ArrayVec` in-place so that `len` is equal to `new_len`.
+    /// Resizes the `ArrayVec` in-place so that its length becomes `new_len`.
     ///
     /// Does the same thing as `try_resize`, but panics on error instead of
     /// returning it as a `Result`.
-    pub fn resize(&mut self, new_len: usize, value: A::Item) {
+    ///
+    /// **Panics** if the corresponding `try_resize` would return an error.
+    pub fn resize(&mut self, new_len: usize, value: A::Item) where A::Item: Clone {
         self.try_resize(new_len, value).unwrap()
     }
 
-    /// Resizes the `ArrayVec` in-place so that `len` is equal to `new_len`.
+    /// Resizes the `ArrayVec` in-place so that its length becomes `new_len`.
     ///
-    /// If `new_len` is greater than `len`, the `ArrayVec` is extended by the
-    /// difference, with each additional slot filled with `value`. If `new_len`
-    /// is less than `len`, the `ArrayVec` is simply truncated.
+    /// If `new_len` is greater than the vector's current length, the
+    /// `ArrayVec` is extended by the difference, with each additional slot
+    /// filled with `value`. If `new_len` is less than the vector's current
+    /// length, the `ArrayVec` is simply truncated.
     ///
     /// This method requires `Clone` to clone the passed value. If you'd rather
     /// create a value with `Default` instead, see `try_resize_default`.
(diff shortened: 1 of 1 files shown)


bluss:
This duplicates the set Len on drop-ish struct. It can be refactored, can't it?

bluss:
Thanks a lot for working on this. To clarify from before, it looks like this duplicates instead of factors out the logic inside extend and its ScopeExitGuard (which has the same function as SetLenOnDrop). It's best to not duplicate it.

What's the reason for the extra methods and types? Maybe the existing `.extend()` can handle it well?

tbu-:
@bluss I basically copied it from stdlib's implementation. `extend` is a little worse here, because it has to clone the elements one time too often.

src/lib.rs
```diff
@@ -597,6 +597,190 @@ impl<A: Array> ArrayVec<A> {
     }
 }
 
+impl<A: Array> ArrayVec<A>
+    where A::Item: Clone,
+{
```
bluss: Style wise I'd prefer a single `impl` block. The where clause can go on each method fine. Split `impl` blocks feels archaic.

src/lib.rs
```diff
@@ -597,6 +597,190 @@ impl<A: Array> ArrayVec<A> {
...
+    /// Resizes the `ArrayVec` in-place so that `len` is equal to `new_len`.
+    ///
+    /// Does the same thing as `try_resize`, but panics on error instead of
+    /// returning it as a `Result`.
```
bluss: Documentation for panic should be consistent and visible. Other methods use this style, panic the first word of the line and bold:

```rust
/// ***Panics*** if the `index` is out of bounds.
```

src/lib.rs
```diff
@@ -597,6 +597,190 @@ impl<A: Array> ArrayVec<A> {
...
+    pub fn resize(&mut self, new_len: usize, value: A::Item) {
+        self.try_resize(new_len, value).unwrap()
+    }
+
+    /// Resizes the `ArrayVec` in-place so that `len` is equal to `new_len`.
```
bluss: Rust's own formulation in their docs "so that `len` is equal" is awkward. Sounds better to just say "so that its length becomes `new_len`" or similar.

src/lib.rs
```diff
@@ -597,6 +597,190 @@ impl<A: Array> ArrayVec<A> {
...
+    ///
+    /// If `new_len` is greater than `len`, the `ArrayVec` is extended by the
```
bluss: I think we should say length instead of `len`.

src/lib.rs
```diff
@@ -597,6 +597,190 @@ impl<A: Array> ArrayVec<A> {
...
+
+impl<A: Array> ArrayVec<A> {
+    /// Extend the vector by `n` values, using the given generator.
+    ///
+    /// Doesn't extend the vector and returns an error if the number of added
+    /// elements exceed the capacity of the underlying array.
+    fn extend_with<E: ExtendWith<A::Item>>(&mut self, n: usize, value: E)
+        -> Result<(), CapacityError>
+    {
+        if self.len() + n > self.capacity() {
```
bluss: This addition feels wrong, as if it could wrap around. But it's a private function and all callers are without problem.

tbu-:
Done I think.

bluss:
Cool. Please have some patience. I'll make a new suggestion :-)

src/lib.rs
```diff
@@ -595,6 +596,237 @@ impl<A: Array> ArrayVec<A> {
...
+        0
+    }
+}
+
+impl<A: Array> ArrayVec<A> {
+    /// Extend the vector by values specified , using the given generator.
+    ///
+    /// Doesn't extend the vector and returns an error if the number of added
+    /// elements exceed the capacity of the underlying array.
+    fn extend_with<E>(&mut self, values: E) -> Result<(), CapacityError>
```
bluss: This should just use the existing `extend` I think.

src/lib.rs
```diff
@@ -595,6 +596,237 @@ impl<A: Array> ArrayVec<A> {
...
+    type Item = T;
+    fn next(&mut self) -> Option<T> {
+        if self.0 == 0 {
+            return None;
+        }
+        self.0 -= 1;
+        Some(if self.0 > 0 {
+            self.1.as_ref().unwrap().clone()
+        } else {
+            self.1.take().unwrap()
```
bluss: These `.unwrap` are avoidable, aren't they? The Some/None-ness can be used to guard the end of the iteration.

src/lib.rs
```diff
@@ -595,6 +596,237 @@ impl<A: Array> ArrayVec<A> {
...
+        })
+    }
+}
+impl<T: Clone> ExtendIterator for ExtendElement<T> {
+    fn needed_capacity(&self) -> usize {
+        self.0
+    }
+}
+
+struct ExtendDefault<T: Default>(usize, PhantomData<T>);
```
bluss: The ExtendDefault iterator is not needed, it can be passed as a regular iterator inline, maybe something like `(0..n).map(|_| X::default())`

src/lib.rs
```diff
@@ -595,6 +596,237 @@ impl<A: Array> ArrayVec<A> {
...
+        Some(Default::default())
+    }
+}
+impl<T: Default> ExtendIterator for ExtendDefault<T> {
+    fn needed_capacity(&self) -> usize {
+        self.0
+    }
+}
+
+struct ExtendIter<I: Iterator>(I);
```
bluss: This trivial wrapper is not needed.

src/lib.rs
bluss: One could spy at itertools' repeat_n, but I'm not sure if that spoils the fun.

bluss: or indeed if it spoils the performance.

bluss:
I'll pick this up at some point

Issue #83, opened 2017-11-03
compile error with avr-rust


wez:
```
error[E0119]: conflicting implementations of trait `array::Array` for type `[_; 0]`:
  --> /Users/wez/.cargo/registry/src/github.com-1ecc6299db9ec823/arrayvec-0.4.5/src/array.rs:65:9
   |
65 |            unsafe impl<T> Array for [T; $len] {
   |   _________^
   |  |_________|
   | ||
66 | ||             type Item = T;
67 | ||             type Index = $index_type;
68 | ||             #[inline(always)]
...  ||
73 | ||             fn capacity() -> usize { $len }
74 | ||         }
   | ||         ^
   | ||_________|
   | |__________first implementation here
   |            conflicting implementation for `[_; 0]`
...
90 |    fix_array_impl_recursive!(u32, 1 << 16,);
   |    ----------------------------------------- in this macro invocation

error: aborting due to previous error

error: Could not compile `arrayvec`.
```

commenting out the 32-bit index case in `array.rs:90` resolves this compilation error.

I'm not sure of the best way to resolve this otherwise I'd just submit a PR to do it :-/

bluss:
Interesting bug! It's def our fault and we can fix it.

wez:
FWIW, it takes a couple of hours to build `avr-rust` (on a macbook pro; you may have a more beefy environment that goes faster!) so I'd be happy to try some things locally if that will save you some time.

bluss:
I think that we just have to make the 1 << 16 implementation conditional. We can't work with so big arrays on 16-bit platforms -- they can't exist there! We need a 17-bit integer (or larger) to represent the arrayvec in its current implementation. I'm considering patching both the 0.3 and the current 0.4 branch, since I have the time to do it.

cc @tbu-

bluss:
@wez I feel relatively confident it's going to work. I want to make something like the following change, which you can try if you have time:

```
diff --git src/array.rs src/array.rs
index 855c117..8276ca9 100644
--- src/array.rs
+++ src/array.rs
@@ -105,5 +105,7 @@ fix_array_impl_recursive!(u8, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14,
                           16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31,
                           32, 40, 48, 56, 64, 72, 96, 128, 160, 192, 224,);
[...]

PR #84, opened 2017-11-03
Fix 1 << 16 size array impl: only possible on bigger than 16-bit (0.3 branch)

bluss:
Fixes, in the 0.3 branch, bug #83 

And two 0.3 branch fixes picked from master

- Correct license files
- Remove broken test


bluss:
Picked the failing test fix from master. The test is UB, so it was broken.

PR #90, opened 2018-01-13
Support Rust 1.13.0

jeehoonkang:
Hi!  First of all, thank you for maintaining this wonderful library!

Currently, `arrayvec` is used by [`crossbeam-epoch`](https://github.com/crossbeam-rs/crossbeam-epoch), which is a concurrency library.  We Crossbeam developers aim to [push it](https://github.com/rayon-rs/rayon/pull/480) to [`rayon`](https://github.com/rayon-rs/rayon), which is a parallelism library.

But `rayon` should support Rust 1.12, and `arrayvec` currently doesn't.  May I ask if you could merge this PR so that `arrayvec` can be compiled in Rust 1.12.1, and in turn `rayon` can depend on `crossbeam-epoch`?

(I agree that the original code is clearer and more idiomatic..  I'm sorry for asking an ugly change.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ env:
   - FEATURES='serde-1'
 matrix:
   include:
-    - rust: 1.14.0
+    - rust: 1.13.0
     - rust: stable
       env:
         - NODEFAULT=1
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,7 +21,7 @@
 //!
 //! ## Rust Version
 //!
-//! This version of arrayvec requires Rust 1.14 or later.
+//! This version of arrayvec requires Rust 1.13 or later.
 //!
 #![doc(html_root_url="https://docs.rs/arrayvec/0.4/")]
 #![cfg_attr(not(feature="std"), no_std)]


bluss:
Please update the .travis.yml -- we don't promise a version compat that we don't test.

bluss:
We'll do our best here. That said, there is a big chance that we will jump to Rust 1.20 and ManuallyDrop in the next version.

bluss:
You must also update:

https://github.com/bluss/arrayvec/blob/master/src/lib.rs#L24

jeehoonkang:
Thanks!  I updated the doc, and added a Travis build target.

I couldn't `cargo +1.12.1 test` this branch; it seems `serde` needs Rust 1.13.  Instead, I succeeded in `cargo +1.13.0 test` this branch.  I'm fine with this version, since we have a chance that the rayon maintainers will bear with bumping the minimum Rust version to 1.13: https://github.com/rayon-rs/rayon/pull/480#issuecomment-357534578

bluss:
:smile: and no code changes needed now. Thanks!

PR #95, opened 2018-02-22
Add ArrayVec::drain_filter()

niklasf:
This adds `drain_filter()` with an implementation similar to [Vec's `drain_filter()`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain_filter). Can also be used to [simplify `retain()`](https://github.com/rust-lang/rust/pull/48065).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -483,22 +483,7 @@ impl<A: Array> ArrayVec<A> {
     pub fn retain<F>(&mut self, mut f: F)
         where F: FnMut(&mut A::Item) -> bool
     {
-        let len = self.len();
-        let mut del = 0;
-        {
-            let v = &mut **self;
-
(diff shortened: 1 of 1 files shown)


bluss:
Thanks, this looks good!  unstable drain_filter in Rust seems a bit volatile, so I'd like to avoid merging this until we are confident it matches the upstream shape of the API. The alternative would be to merge it, but remove it in arrayvec 1.0.

https://github.com/rust-lang/rust/issues/43244

niklasf:
Makes sense. Let's put it on hold.

bluss:
Hm, still no movement on the tracking issue.

niklasf:
Possibly relevant, one day: https://github.com/rust-lang/rust/pull/61224

PR #97, opened 2018-03-25
Implement a careful “MaybeUninit” using ManuallyDrop

bluss:
This is a cautious version of MaybeUninit, since we don't have one in
libstd, based on ManuallyDrop.

This moves ArrayVec to a nice, no size overhead implementation by
default. We use Rust version sniffing (build script) to automatically
use this for new enough Rust versions.

This doesn't kill nodrop unfortunately, it still remains as the
fallback.

### DIFF
--- a/src/maybe_uninit.rs
+++ b/src/maybe_uninit.rs
@@ -1,0 +1,34 @@
+
+
+use std::mem::ManuallyDrop;
+use std::mem::uninitialized;
+
+/// A combination of ManuallyDrop and “maybe uninitialized”;
+/// this wraps a value that can be wholly or partially uninitialized;
+/// it also has no drop regardless of the type of T.
+pub struct MaybeUninit<T>(ManuallyDrop<T>);
+
+impl<T> MaybeUninit<T> {
+    /// Create a new MaybeUninit with uninitialized interior
+    pub unsafe fn uninitialized() -> Self {
+        Self::from(uninitialized())
+    }
+
+    /// Create a new MaybeUninit from the value `v`.
+    pub fn from(v: T) -> Self {
+        MaybeUninit(ManuallyDrop::new(v))
+    }
+
+    /// Return a raw pointer to the interior
+    pub fn ptr(&self) -> *const T {
+        (&self.0) as *const ManuallyDrop<_> as *const T
+    }
+
+    /// Return a raw pointer to the interior (mutable)
+    pub fn ptr_mut(&mut self) -> *mut T {
+        (&mut self.0) as *mut ManuallyDrop<_> as *mut T
+    }
+}
+
+
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ env:
 matrix:
   include:
     - rust: 1.13.0
+    - rust: 1.19.0
+    - rust: 1.20.0
     - rust: stable
       env:
         - NODEFAULT=1
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,9 @@ repository = "https://github.com/bluss/arrayvec"
 keywords = ["stack", "vector", "array", "data-structure", "no_std"]
 categories = ["data-structures", "no-std"]
 
+[build-dependencies]
+version_check = "0.1"
+
 [dependencies]
 nodrop = { version = "0.1.12", path = "nodrop", default-features = false }
 
@@ -37,8 +40,10 @@ harness = false
 [features]
 default = ["std"]
 std = []
-use_union = []
 serde-1 = ["serde"]
+
+# has no effect
+use_union = []
 
 [package.metadata.docs.rs]
 features = ["serde-1"]
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,13 @@
+
+//!
+//! This build script detects if we have new enough Rust
+//!
+
+extern crate version_check;
+
+fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
+    if let Some((true, _)) = version_check::is_min_version("1.20.0") {
+        println!("cargo:rustc-cfg=has_manuallydrop");
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,14 +7,6 @@
 //!   - Optional, enabled by default
 //!   - Use libstd; disable to use `no_std` instead.
 //!
-//! - `use_union`
-//!   - Optional
-//!   - Requires Rust nightly channel
-//!   - Experimental: This flag uses nightly so it *may break* unexpectedly
-//!     at some point; since it doesn't change API this flag may also change
-//!     to do nothing in the future.
-//!   - Use the unstable feature untagged unions for the internal implementation,
-//!     which may have reduced space overhead
 //! - `serde-1`
 //!   - Optional
 //!   - Enable serialization for ArrayVec and ArrayString using serde 1.0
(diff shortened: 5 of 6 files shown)


Thomasdezeeuw:
I really like the design of [RFC #1892](https://github.com/rust-lang/rfcs/pull/1892), which defines a new type `MaybeUninit`:

```rust
union MaybeUninit<T> {
    uninit: (),
    value: T,
}
```

I think that would be a great fit for `ArrayVec` use case, something like this:

```rust
pub struct ArrayVec<A: Array, T> {
    data: A<Item = MaybeUninit<T>>,
    len: A::Index,
}
```

RalfJung:
Hm, notice that for `ManuallyDrop<T>` we will likely want to decide that it always has to contain at least a "bit-valid" `T`. That's required for unsized `ManuallyDrop` in combination with thin dyn types. So, the `ManuallyDrop` here is likely not correct.

Is it possible to have a `union` of a `ManuallyDrop<T>` and a `()`? That would be safe, I think.

bluss:
Thanks for the pointers!

bluss:
@RalfJung I'll do the best I can with stable Rust here and insert MaybeUninit as soon as it is stable (with a version check).  A union is not possible since we want to have non-Copy `T`:

```rust
error[E0658]: unions with non-`Copy` fields are unstable (see issue #32836)
  --> src/maybe_uninit.rs:15:1
   |
15 | / union Test<T> {
16 | |     empty: (),
17 | |     value: ManuallyDrop<T>,
18 | | }
   | |_^
   |
   = help: add #![feature(untagged_unions)] to the crate attributes to enable
```

bluss:
@RalfJung It's long ago now that I discussed this with @arielb1, and we came into discussion about the difference between "moved from" (like `ptr::read` or ManuallyDrop::drop would leave the interior. [Put in nuance if those are different]) and wholly uninitialized.

Current `ManuallDrop::drop` basically says it leaves the interior uninitialized. So in that sense this type can already contain this.

bluss:
another problem from ArrayString: it is `Copy`, and we'd like to be able to Copy it while the array is partially or wholly uninitialized.

src/maybe_uninit.rs
```diff
@@ -0,0 +1,105 @@
...
+/// it also has no drop regardless of the type of T.
+///
+/// This is a stop-gap solution until MaybeUninit is stable in Rust's std.
+#[repr(C)]
+pub struct MaybeUninit<T>(ManuallyDrop<T>);
+
+impl<T> MaybeUninit<T> {
+    /// Create a new MaybeUninit with uninitialized interior
+    pub unsafe fn uninitialized() -> Self {
+        Self::from(uninitialized())
```
RalfJung: I'm afraid this is not correct -- it is as incorrect as just plain `mem::uninitialized`.

Using terminology from [my blog post](https://www.ralfj.de/blog/2018/08/22/two-kinds-of-invariants.html), the issue with `mem::uninitialized` is that it returns data that violates the validity invariant. While `ManuallyDrop` can handle data that is already dropped, it does not have any effect on the validity invariant: `ManuallyDrop<bool>` may only be `0x0` or `0x1`, and hence 
```rust
let x: MaybeUninit<bool> = MaybeUninit::uninitialized();
```
with your type is UB.

For non-`Copy` types, there is currently no way to do this. We should really work towards solving the remaining open questions in https://github.com/rust-lang/rust/issues/53491!

bluss: FWIW; I know this is not correct, this is the current best effort with stable Rust like the PR says.

The current state of arrayvec as it is released, can't be said to be any better than this, I'm sure?

bluss: @RalfJung With MaybeUninit, what's the importance of using `MaybeUninit<[T; 10]>` vs `[MaybeUninit<T>; 10]` ? The latter I don't even have clear picture of how to initialize, when we start with `T` values that are not `Copy`.

RalfJung: > The current state of arrayvec as it is released, can't be said to be any better than this, I'm sure?

I do not know that state. This is pretty much equivalent to using `mem::uninitialized` directly.

> it uses mem::uninitialized() to initialize a whole value of [MaybeUninit<T>; 10].

oO I had no idea.^^ Where can I find its sources?

Time that we get `mem::uninitialized` deprecated so that these things are rejected by `-D warnings`.

But given that `MaybeUninit` is a union, it is actually okay to not initialize it. The problem with `mem::uninitialized` is that it almost always violates the validity invariant of the type it returns; but for `MaybeUninit`, the validity invariant accepts literally any data (including uninitialized data), and hence there is no problem with that particular usage of `mem::uninitialized`.

RalfJung: Looking at `ArrayVec` when it got removed in rust-lang/rust@687cc292fd681be9739dc973acd5eaa5f73a5ce7, it uses `mem::uninitialized` on `ManuallyDrop`, which is *not* okay.

bluss: Ok! By the way I'm following the stacked borrows discussion and unsafe code guidelines now. It will have a big impact on ndarray, I believe. Just like the rest of the ecosystem we are gradually getting a clearer picture of what we can do in unsafe code and can dot our i's and cross the t's here.

I would go with using `MaybeUninit<[T; 10]>` and offsetting into it, because the other way around leads to code with more transmutes when getting initialized `[T; 10]` wrapped into the ArrayVec.

Given this, it sounds like it is ok to use `mem::uninitialized` with `[MaybeUninit<T>; 10]`, even if the array is the outer type. That's the only answer to how to set a value to such a field that I can see (or equivalent union trick).

eddyb:
> A `union` is not possible since we want to have non-`Copy` `T`:

I thought we were going to allow `ManuallyDrop<T>`, *alongside* any `Copy` types, as the property we're interested in isn't *really* `Copy`, but "needs no drop".

Did we not stabilize that? If so, why?

RalfJung:
> Did we not stabilize that? If so, why?

We accepted the RFC. It hasn't been implemented yet, though.

bluss:
This is confirmed broken, it's something that has changed since this was first prototyped using Rust 1.20. 

Now that `ManuallyDrop<T>` switched to struct, its interior is no longer exempt from being used enum layout optimization, and this makes an arrayvec based on ManuallyDrop noticeably unsound, and smallvec has the same problem.

Testing the current PR:

```rust
#[test]
fn test_still_works_with_option_arrayvec() {
    type RefArray = ArrayVec<[&'static i32; 2]>;
    let array = Some(RefArray::new());
    assert!(array.is_some());
    println!("{:?}", array);
}
```

The test passes fails in debug mode, it passes in release mode but the output is "None", which is incorrect.

See also servo/rust-smallvec/issues/126 for smallvec

RalfJung:
The fact that `ManuallyDrop` was exempt from layout optimizations was an implementation limitation, it was never a guarantee.

RalfJung:
Agreed, it is a problem. However, [the relevant change](https://github.com/rust-lang/rust/pull/52711) happened several months ago and got released on stable with 1.29, 10 weeks ago.

I think the main problem here is people relying on things that just "happen to be the case". I am not sure how to prevent this, other than repeating all the time "don't do that", and working on stabilizing APIs that let people better express their intent.

RalfJung:
IOW, we should work hard towards stabilizing `MaybeUninit` :D

> And the original crate was written long before anyone could answer what was and wasn't allowed, it was just unclear.

Relying on unclear things is exactly what I meant.  It's dangerous.

But I also appreciate that so many things are unclear still that it can be hard to avoid.

bluss:
@RalfJung :smile:  https://github.com/rust-lang/rust/pull/56466  The old sins are disappearing

bluss:
See PR #114 for another pragmatic mitigation—fighting the fire where we can—conditional use of unions and a good solution (only on nightly!)

PR #98, opened 2018-03-25
Use drop_in_place for truncate and clear (and drop)

bluss:
This should perform better in both release and debug mode.

NOTE: This is significant because it changes the drop order of the
elements, and how we handle panicking destructors. If just one of the
destructors panic during ArrayVec drop, clear or truncate, the rest of
the elements should still drop. If we encounter another panic during
that process, however, Rust will abort as usual for panic during
unwinding.

The new tests for panic during drop/truncate don't pass unless
the `drop_in_place` change is applied. Thus we are more correct
when it comes to dropping all elements with this.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -459,12 +459,18 @@ impl<A: Array> ArrayVec<A> {
     /// assert_eq!(&array[..], &[1, 2, 3]);
     /// ```
     pub fn truncate(&mut self, len: usize) {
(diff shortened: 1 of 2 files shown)


bluss:
@Thomasdezeeuw Sorry I was away. Out of curiosity, why is the change needed?

bluss:
I see. The loop should have optimized out in that case, but maybe it never did! That's kind of an oversight.

PR #102, opened 2018-08-31
Fix undefined behavior in DerefMut of ArrayString

gnzlbg:
The representations of `&mut[T]` is unspecified for all `T` and the representation of `str` is unspecified as well so the `transmute` here is undefined behavior.


gnzlbg:
This is a breaking change in that the minimum supported Rust version has to be upgraded to Rust 1.20.0.

bluss:
Thanks, solid change

PR #104, opened 2018-10-14
Implement Clone, Debug for IntoIter

clarcharr:


src/lib.rs
```diff
@@ -748,6 +748,18 @@ impl<A: Array> Drop for IntoIter<A> {
     }
 }
 
+impl<A: Array> Clone for IntoIter<A> where A::Item: Clone {
+    fn clone(&self) -> IntoIter<A> {
+        IntoIter { index: self.index, v: self.v.clone() }
```
bluss: We have potentially moved out from v, so we can't call clone here

clarcharr: Oh, whoops, thank you. Updated to a `.collect().into_iter()`

clarcharr:
Fixed that, and added a rather naïve test to make sure that cloning doesn't drop anything more than it should.

bluss:
Cool. Can we get that without the reformatting? Thanks. Sorry that I'll be away from maintaining a lot. So it's not a hurry.

clarcharr:
Done! And it's totally fine; life happens.

bluss:
Thanks!

PR #111, opened 2018-11-25
Fix just a few stacked borrow issues, and improve extend performance

bluss:
Extend: This simplification -- borrowing self.len instead of self, leads to
an improvement in the extend_from_slice benchmark.

It's also guided by the discussion of stacked borrows; the old code
would be invalid, because the whole self is borrowed while ptr is derived from
self.

In truncate, don't access self while holding raw pointer derived from self

Again, stacked borrows model makes the `self.set_len()` call illegal
because we are holding (and are going to use) another raw pointer
derived from self, `tail`.

bench comparison for extend. Caveat might be that optimization can also be influenced by how the code can be optimized in the benchmark.

```
 name                  63 ns/iter       62 ns/iter       diff ns/iter   diff % 
 extend_with_constant  290 (1765 MB/s)  296 (1729 MB/s)             6    2.07% 
 extend_with_range     288 (1777 MB/s)  287 (1783 MB/s)            -1   -0.35% 
 extend_with_slice     828 (618 MB/s)   218 (2348 MB/s)          -610  -73.67%
```

### DIFF
--- a/benches/extend.rs
+++ b/benches/extend.rs
@@ -5,6 +5,7 @@ extern crate arrayvec;
 use arrayvec::ArrayVec;
 
 use bencher::Bencher;
+use bencher::black_box;
 
 fn extend_with_constant(b: &mut Bencher) {
     let mut v = ArrayVec::<[u8; 512]>::new();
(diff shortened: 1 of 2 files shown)


bluss:
There is bound to be more issues to fix than this. cc #97

PR #112, opened 2018-11-25
Improve .extend() performance (?)

bluss:
cc #101 

I'm experimenting with different formulations. Both the existing and this pr's implementations can compile into a memcpy, like they should, if the input is a cloned slice iterator. The main difficulty seems to be to make a good benchmark. Without "black_box"es, the benchmarks compile out (and that's normally a good sign in itself, the code is then transparent to the optimizer) and with too many black box calls, the optimizations are disabled.

### DIFF
--- a/benches/extend.rs
+++ b/benches/extend.rs
@@ -2,6 +2,8 @@
 extern crate arrayvec;
 #[macro_use] extern crate bencher;
 
+use std::io::Write;
+
 use arrayvec::ArrayVec;
 
 use bencher::Bencher;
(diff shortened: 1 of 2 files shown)


bluss:
Not entirely happy with these benchmarks either, see code in the PR, but they seem fair(? please review)

```
 name                  63 ns/iter       62 ns/iter       diff ns/iter   diff % 
 extend_with_constant  294 (1741 MB/s)  1 (512000 MB/s)          -293  -99.66% 
 extend_with_range     426 (1201 MB/s)  289 (1771 MB/s)          -137  -32.16% 
 extend_with_slice     424 (1207 MB/s)  13 (39384 MB/s)          -411  -96.93% 
 extend_with_write     13 (39384 MB/s)  13 (39384 MB/s)             0    0.00%
```

obviously when extend_with_constant optimizes out it doesn't tell us so much, except that the new extend code is somehow more transparent to the compiler than the old.

bluss:
Comparison with `try_extend_from_slice` shows that they both compile to memcpy:

```
test extend_from_slice    ... bench:          14 ns/iter (+/- 1) = 36571 MB/s
test extend_with_slice    ... bench:          13 ns/iter (+/- 1) = 39384 MB/s
```

`extend_with_slice` is the regular `extend()` used with a slice iterator (benchmark is in the PR).

PR #114, opened 2018-12-15
Implement a "MaybeUninit" and use it conditionally (0.4.x version)

### LINKED ISSUE #86: update docs around use_union
The docs still say that `use_union` requires the nightly channel, but it looks like it no longer does? (Though it also sounds like there are some [safety questions](https://github.com/bluss/arrayvec/issues/62) still open?)

bluss:
This change is for the 0.4.x release series of arrayvec,
not master.

**Implement a "MaybeUninit" and use it conditionally**

Use a build script to detect if we can use MaybeUninit or NoDrop.
Enabling unstable features automatically is not ideal, but since it's
a soundness issue we should do it.

Use a MaybeUninit-like union on nightly when we can. We use a feature
detection script in build.rs, so that we also go back to the fallback if
the unstable feature changes in an unexpected way.

We need to continue to use NoDrop for best working stable
implementation, but we eagerly use our union solution where we can,
currently only in nightlies.

Rustc feature probe code written by @cuviper, taken from num-bigint.

**Remove use of uninitialized in ArrayString**

We can't fix this properly (MaybeUninit with a union) until we change
the user visible API (we need to require that A: Copy.

As a temporary solution for arrayvec version 0.4.*, we use zeroed to
initialize an array of bytes, instead of using uninitialized. This may
have a negative performance impact, but the fix is to upgrade to future
arrayvec 0.5.

This also changes the `use_union` feature to do nothing. This is
allowed, as has been indicated in the feature description in docs.

There are no user visible API changes but the size of ArrayVec types may
change due to the enabling of the MaybeUninit feature.

Closes #86 (now obsolete)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,13 +19,16 @@ matrix:
     - rust: nightly
       env:
         - NODEFAULT=1
+        - ARRAYVECTEST_ENSURE_UNION=1
     - rust: nightly
       env:
       - NODROP_FEATURES='use_needs_drop'
+      - ARRAYVECTEST_ENSURE_UNION=1
     - rust: nightly
       env:
       - FEATURES='serde use_union'
       - NODROP_FEATURES='use_union'
+      - ARRAYVECTEST_ENSURE_UNION=1
 branches:
   only:
     - master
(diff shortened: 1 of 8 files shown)


bluss:
This will be merged into master after 0.4, and we'll make a release with similar code for 0.5.x release series.

Thougths on updating 0.4.x? It's quite widely used, so it would be good to put a sounder implementation out there, if only on nightly. We're doing the best we can with stable.

PR #117, opened 2018-12-23
Implement FromStr for ArrayString

despawnerer:
This is very useful for generic code that may want to parse
input strings into arbitrary other types, such as ArrayString.

Limitations of the FromStr trait don't allow us to keep the original string slice
inside the CapacityError, unfortunately.


bluss:
Thanks!

PR #123, opened 2019-06-12
Implement quickcheck::Arbitrary trait for ArrayVec

podhrmic:
When you use `ArrayVec` and `quickcheck`, it is really handy to have `Arbitrary` trait implemented for `ArrayVec` - trivial (and optional) change, but IMHO useful.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,6 +28,12 @@ extern crate serde;
 #[cfg(not(feature="std"))]
 extern crate core as std;
 
+#[cfg(feature="quickcheck-1")]
+extern crate rand;
+
+#[cfg(feature="quickcheck-1")]
+extern crate quickcheck;
+
 #[cfg(not(has_manually_drop_in_union))]
 extern crate nodrop;
 
@@ -63,6 +69,12 @@ use maybe_uninit_copy::MaybeUninitCopy;
 
 #[cfg(feature="serde-1")]
 use serde::{Serialize, Deserialize, Serializer, Deserializer};
+
+#[cfg(feature="quickcheck-1")]
+use rand::Rng;
+
+#[cfg(feature="quickcheck-1")]
+use quickcheck::{Arbitrary,Gen};
 
 mod array;
 mod array_string;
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -987,6 +999,23 @@ impl<A: Array> Clone for ArrayVec<A>
...
 
+#[cfg(feature="quickcheck-1")]
+impl<A: Array + Send + Clone + 'static > Arbitrary for ArrayVec<A>
+where <A as array::Array>::Item: std::clone::Clone + Arbitrary,
+<A as array::Array>::Index: std::marker::Send
+{
+    fn arbitrary<G: Gen>(g: &mut G) -> Self {
+        use std::cmp::{min, max};
+        let mut v = Self::default();
+        let size = g.gen_range(0, max(1, min(g.size(), v.capacity())));
```
tbu-:
```suggestion
        let size = g.gen_range(0, max(1, min(g.size(), v.capacity() + 1)));
```

Since `gen_range` is exclusive on the right side, `v.capacity() + 1` should be good as well.

src/lib.rs
```diff
@@ -987,6 +999,23 @@ impl<A: Array> Clone for ArrayVec<A>
     }
 }
 
+#[cfg(feature="quickcheck-1")]
+impl<A: Array + Send + Clone + 'static > Arbitrary for ArrayVec<A>
```
tbu-:
```suggestion
impl<A: Array + Send + Clone + 'static> Arbitrary for ArrayVec<A>
```

src/lib.rs
```diff
@@ -987,6 +999,23 @@ impl<A: Array> Clone for ArrayVec<A>
...
+where <A as array::Array>::Item: std::clone::Clone + Arbitrary,
```
tbu-:
```suggestion
    where <A as array::Array>::Item: Clone + Arbitrary,
```

src/lib.rs
```diff
@@ -987,6 +999,23 @@ impl<A: Array> Clone for ArrayVec<A>
...
+<A as array::Array>::Index: std::marker::Send
```
tbu-:
```suggestion
          <A as array::Array>::Index: Send,
```

src/lib.rs
```diff
@@ -987,6 +999,23 @@ impl<A: Array> Clone for ArrayVec<A>
...
+{
+    fn arbitrary<G: Gen>(g: &mut G) -> Self {
+        use std::cmp::{min, max};
+        let mut v = Self::default();
+        let size = g.gen_range(0, max(1, min(g.size(), v.capacity())));
+        (0..size).for_each(
```
tbu-: This is probably more idiomatically implemented with a for loop.

podhrmic: Fixed!

podhrmic:
@tbu- thanks for the feedback!

bluss:
The policy is that arrayvec 1.0 can not depend on pre-1.0 crates. If I run those rules, that means we could accept this, but the feature will be tossed out without regrets when we are ready for arrayvec 1.0.  Which we should be, but since that is moving glacially, that's not a good reason to hold this back. ;)

I'd try to skip the direct rand dependency just for simplicity (we don't have to manually match the rand version quickcheck uses). I also see why the feature name quickcheck-1 was chosen, but for serde-1 it means it's a dependency on the 1.x series of serde.

podhrmic:
@bluss I don't have a strong opinion, but also I won't be able to make any changes in this PR at least for a couple of weeks. So depends what you think is best way forward.

bluss:
Ok. I'd implement it like this, if it works to drop direct rand dependency. https://github.com/bluss/arrayvec/compare/quickcheck-0.8?expand=1 

Still needs tests. Using a dependency rename with the version should be a good strategy.

PR #126, opened 2019-06-24
Implement Extend/FromIterator for ArrayString

### LINKED ISSUE #119: Should ArrayString be Item=char?
Hi,
I want to use ArrayString in `rustc_hex::ToHex`, this trait works on every type that implements FromIterator on Item=char.
But ArrayString is Item=u8.

Is there a reason this isn't Item=char?

clarfon:
Fixes #119. Caveat: takes the most conservative approach and doesn't extract any characters when there are less than 4 bytes left in the array.

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -8,6 +8,7 @@ use std::str;
 use std::str::FromStr;
 use std::str::Utf8Error;
 use std::slice;
+use std::iter;
 
 use array::{Array, ArrayExt};
 use array::Index;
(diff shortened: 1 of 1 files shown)


bluss:
Hey, thanks for working on this. The pain point I've seen here before is exactly how do deal with capacity and error handling, since the two traits are not designed to do that. I think that this solution of leaving up to 3 unused capacity in a byte array, while understandable, is too confusing to use in arrayvec, and would in fact often go counter to the user's wish(?), say if they want to fit 10 ascii letters into 10-capacity array.

clarfon:
So to clarify, should I just close this PR then? Or would it be good to leave open as an option for the future?

bluss:
Yeah, let's close this for now. I think we haven't found good rules for how it should work.

PR #129, opened 2019-09-01
Always use MaybeUninit and stop using nodrop fallback, prepare next release

bluss:
Switch to using MaybeUninit for everything (no more nodrop fallback for older Rust versions).

Use std::mem::MaybeUninit and stop using nodrop as a fallback.
This means we require Rust 1.36.

And prepare release notes partially for the next release.
Update ArrayExt methods in anticipation of next release.


bluss:
Completely sidenote, but arrayvec's test suite passes with experimental const generics (not part of the PR). However one of my arrayvec applications has unexplained crashes that I haven't been able to reduce right now, with the const generics arrayvec, so it's not exactly ready.

c410-f3r:
@bluss 

Do you mean that you are already working with const generic stuff?

bluss:
@c410-f3r It was a 10 line patch to try it, it's not more than that to implement. So yes and no. I've tried it, and evaluated it and the conclusion so far is in the previous comment, no further work right now.

c410-f3r:
@bluss 

Ok, thanks! It is good to know that as I was about to start working with it to create a PR.

bluss:
@c410-f3r on closer look it might be related to MaybeUninit changes — so it's worth trying. And it's easy to drop in const generics by changing the Array trait impls, but I guess we actually want to change the parameterization of ArrayVec itself to actually do it, which is more work.

c410-f3r:
> @c410-f3r on closer look it might be related to MaybeUninit changes — so it's worth trying. And it's easy to drop in const generics by changing the Array trait impls, but I guess we actually want to change the parameterization of ArrayVec itself to actually do it, which is more work.

Great. I will start look into it in the next few days.

> Keep in mind that we can't really merge anything until const generics are stable - maybe there's a good way to have it as a preview unstable feature, but we'll need to use some care. An unstable new Array impl could be an ok middle ground.

I was basically thinking of doing the same thing I did in `smallvec`, which is a new unstable `const_generics` feature controlled by a macro invocation.
Hope I can soon provide a PR to give a better understanding of the overall changes.

bluss:
@c410-f3r Please don't publish the same kind of PR here as the one to smallvec - keep separate changes separate and well contained, and you'll save yourself from heartbreak. We're as likely to be conservative here as there, about what changes we want to incorporate.

PR #133, opened 2019-10-01
gitignore: ignore target as a symlink

mathstuf:


bluss:
Curious where this use case comes from

tbu-:
If you want to move all cargo target dirs globally, just do
```
export CARGO_TARGET_DIR=/var/tmp/cargo
```

mathstuf:
I keep my build trees on a separate partition and wrangled using symlinks.

> If you want to move all cargo target dirs globally

AFAIK, this doesn't work because then all builds of Rust code on my machine is serialized on the lock in the one target directory (I already have issues with ALE competing with my `:mak build` commands in Vim as it is).

bluss:
I would half suggest that you should just use a personal excludes configuration in git for that

mathstuf:
I'm aware of that; I use it for things like my Vim backup files and the like. But this is a file that "every" clone of this ends up having. It's basically a backport of the new behavior of `cargo new`: rust-lang/cargo#4944

PR #140, opened 2019-11-05
impl diesel::deserialize::FromSql for ArrayString

AtsukiTak:
This PR adds support for `diesel`. With `diesel` feature flag, you can query `text` type field without any allocation. This is what I really wanted :)
If this feature is too much for `arrayvec` crate, I would like to create another crate. Thanks.

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -17,6 +17,13 @@ use crate::char::encode_utf8;
 #[cfg(feature="serde")]
 use serde::{Serialize, Deserialize, Serializer, Deserializer};
 
+#[cfg(feature="diesel")]
+use diesel::{
+    deserialize::{FromSql},
+    backend::Backend,
+    FromSqlRow,
+};
+
 use super::MaybeUninit as MaybeUninitCopy;
 
 /// A string with a fixed capacity.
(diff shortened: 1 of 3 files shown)


src/array_string.rs
```diff
@@ -565,3 +573,19 @@ impl<'de, A> Deserialize<'de> for ArrayString<A>
...
+
+#[cfg(feature="diesel")]
+/// Requires crate feature `"diesel"`
+impl<A, ST, DB> FromSql<ST, DB> for ArrayString<A>
+    where A: Array<Item = u8> + Copy,
+          DB: Backend,
+          *const str: FromSql<ST, DB>
+{
+    fn from_sql(bytes: Option<&DB::RawValue>) -> diesel::deserialize::Result<Self> {
+        let str_ptr = <*const str as FromSql<ST, DB>>::from_sql(bytes)?;
```
bluss: What's the argument for this code being correct? What does FromSql do for `*const str` ?

bluss:
Probably arrayvec does not want to have a public dependency on diesel

AtsukiTak:
So sorry for super delay and thank you so much for reviewing!

PR #142, opened 2019-11-08
impl FromIterator<char> for ArrayString

oconnor663:
String includes a number of other impls for e.g. &char and &str, and we
could expand the impls here in a follow-up commit if we want.

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -2,6 +2,7 @@ use std::borrow::Borrow;
 use std::cmp;
 use std::fmt;
 use std::hash::{Hash, Hasher};
+use std::iter::FromIterator;
 use std::ptr;
 use std::ops::{Deref, DerefMut};
 use std::str;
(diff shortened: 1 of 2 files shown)


oconnor663:
This enables you to `.collect()` an `ArrayString` from any iterator of chars. This is a relevant bound for e.g. the [`hex::ToHex`](https://docs.rs/hex/0.4.0/hex/trait.ToHex.html) trait.

clarfon:
This is a duplicate of #126, which was ultimately not merged.

oconnor663:
Oh thanks for the link. One thing this highlights for me is that I misunderstood how `ArrayVec::collect` works. I had assumed it panics if it doesn't have enough capacity, but actually it ignores the rest of the iterator, as in #126.

My main reason for preferring to panic was that it fit my use case :) Namely, [I know exactly how long a certain string is going to be](https://github.com/oconnor663/blake2_simd/blob/0.5.9/blake2s/src/lib.rs#L571-L609), and I've allocated exactly enough space to contain it. If it turned out I didn't have enough space, that would be a bug in my code, and I would want to see a panic telling me to fix it.

Do you find the "ignore the rest of the iterator" use case common? If so, then I don't have any better ideas than what you discussed in #126, and I agree that not implementing this at all might be better than implementing something surprising. But is there any chance a future version of ArrayVec might prefer to switch to panicking behavior for `extend` and `collect`?

clarfon:
I don't think there's going to be a good sense of what's best until we have a clearer idea of what falliable `Extend` and `FromIterator` look like in std. So, for now, this is the behaviour that made the most sense.

PR #143, opened 2019-11-14
Add is_empty method for ArrayVec and ArrayString

nicbn:
As per clippy hint (it's clearer and more specific than comparing `len()` against `0`).

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -71,6 +71,10 @@ impl<A> ArrayString<A>
     #[inline]
     pub fn len(&self) -> usize { self.len.to_usize() }
 
+    /// Returns whether the string is empty.
+    #[inline]
+    pub fn is_empty(&self) -> bool { self.len() == 0 }
+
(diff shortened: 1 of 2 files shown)


clarfon:
This is already provided by the `Deref` impl. Honestly, the inherent `len` shouldn't exist, and it only exists on `Vec` due to the fact that it would be backwards-incompatible.

bluss:
Super minor point but `len` should exist here as on `Vec`, because we want to be able to access the fundamental properties (length and pointer to backing store) without going through deref (going through a potentially temporarily invalid or uninit `&[T]`), for maximal flexibility and correctness for unsafe code. This was the reason `Vec` was complimented with `Vec::as_ptr` in Rust 1.37 and ArrayVec follows that idea.

I'm not a fan of the duplication, but the same reasoning makes it valid to add `is_empty` because it is a variation of `len`.

bluss:
Merged because of what was previously said (I'm not a fan of the clippy hint - is_empty is clearly already provided).

Issue #103, opened 2018-09-22
Add a way to append another array


jD91mZM2:
I'm not sure, but I suspect having an `append` function that consumes another arrayvec and `ptr::copy`s over all elements at once would be faster than using `extend`ing an iterator which `ptr::write`s per-item.

bluss:
Ideally extend shouldn't ptr::write per item, it should optimize better than that, but I suspect the IntoIter implementation is putting limits on how well that can be done.

PR #158, opened 2020-02-13
append and try_append

jmc718:
Added append and try_append with tests to implement https://github.com/bluss/arrayvec/issues/103

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -651,6 +651,33 @@ impl<A: Array> ArrayVec<A> {
         mem::forget(self);
     }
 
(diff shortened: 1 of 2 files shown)


segeljakt:
Any update on this? @bluss

bluss:
Can we do this - don't provide append because it doesn't have a clear error handling story. Users can use extend if they want to do this exact operation already.

try_append is good. I think it should handle both appending arrays and arrayvecs, if possible

burdges:
I'd think `try_extend` gives a better name.

PR #161, opened 2020-06-30
Add unstable-const-fn feature to make new() functions const.

m-ou-se:
This allows for `ArrayVec`s in static variables, as those need to be const initialized.

This is especially useful in embedded software, where global variables are more common and dynamic memory allocation is often avoided.

E.g.

```rust
static RX_BUFFER: Mutex<ArrayVec<[u8; 128]>> = Mutex::new(ArrayVec::new());
```

Unfortunately, this requires a nigthly feature, as generic `const fn`s are only stable for generic parameters with no bounds. Here, the `A: Array` bound is necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,9 @@ matrix:
     - rust: nightly
       env:
       - FEATURES='array-sizes-33-128 array-sizes-129-255'
+    - rust: nightly
+      env:
+      - FEATURES='unstable-const-fn'
     - name: "miri"
       script: sh ci/miri.sh
 branches:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ harness = false
 [features]
 default = ["std"]
 std = []
+unstable-const-fn = []
 
 array-sizes-33-128 = []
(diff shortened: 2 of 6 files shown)


bluss:
Nice that this can be done. How far are we from that this is stable? The trait bounds are a problem aren't they?

I'm not sure we can commit to having unstable features in the crate, but if we could the following remains:

- I'd suggest the feature name `unstable-const-fn`
- The crate feature must be documented (as unstable) at the top of lib.rs with the other features.
- The pull request updated with a description

🙂

m-ou-se:
> How far are we from that this is stable?

Not sure.

> The trait bounds are a problem aren't they?

Yes.

> * I'd suggest the feature name `unstable-const-fn`
> * The crate feature must be documented (as unstable) at the top of lib.rs with the other features.
> * The pull request updated with a description

Done.

m-ou-se:
Can this be merged? Or is there something else that needs to happen first?

bluss:
Yes, why not. Thanks a lot! :)

With my spotty attention, it is fraught to have unstable features in the crate, but it is opt-in.

PR #166, opened 2020-10-21
Use `add` instead of `offset` so we don't need to cast to isize.

hbina:
Signed-off-by: Hanif Bin Ariffin <hanif.ariffin.4326@gmail.com>


bluss:
There are lots of formatting changes in your pr, those don't seem intentional

hbina:
oops, I just automatically rustfmt'ed it

bluss:
Thankss!

PR #167, opened 2020-10-21
Add safety note to the some unsafe functions.

hbina:
Mostly just moving comment and adding some additional informations.

Signed-off-by: Hanif Bin Ariffin <hanif.ariffin.4326@gmail.com>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -516,11 +516,16 @@ impl<A: Array> ArrayVec<A> {
 
     /// Set the vector’s length without dropping or moving out elements
     ///
-    /// This method is `unsafe` because it changes the notion of the
-    /// number of “valid” elements in the vector. Use with care.
     ///
     /// This method uses *debug assertions* to check that `length` is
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -516,11 +516,16 @@ impl<A: Array> ArrayVec<A> {
...
     ///
     /// This method uses *debug assertions* to check that `length` is
     /// not greater than the capacity.
+    ///
+    /// # Safety
+    ///
+    /// This method is `unsafe` because it changes the notion of the
+    /// number of “valid” elements in the vector. This new value
+    /// might not correspond to the actual number of elements in the
+    /// vector which many functions rely to be true. Use with care.
```
bluss, line 528: I wasn't able to just merge this since the way it's written here makes it sound as though it's ok to set the length to a value that does not correspond to the number of initialized elements in the vector. That's not really true: length must be set to a correct value, but the caller is responsible for checking the value (thus the method is unsafe).

PR #170, opened 2020-12-09
Impl TryFrom<fmt::Arguments<'a>> for ArrayString

c410-f3r:
Shortcut for

```rust
use core::fmt::Write;
let mut v = Self::new();
v.write_fmt(format_args!("Something -> {}", 123)).map_err(|e| CapacityError::new(e))?;

...
```

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -594,3 +594,17 @@ where
         Ok(v)
     }
 }
+
+impl<'a, A> TryFrom<fmt::Arguments<'a>> for ArrayString<A>
+where
+    A: Array<Item = u8> + Copy
+{
+    type Error = CapacityError<fmt::Error>;
+
+    fn try_from(f: fmt::Arguments<'a>) -> Result<Self, Self::Error> {
+        use fmt::Write;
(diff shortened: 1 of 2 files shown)


bluss:
Could you add a test for this method?

src/array_string.rs
```diff
@@ -594,3 +594,17 @@ where
...
+where
+    A: Array<Item = u8> + Copy
+{
+    type Error = CapacityError<fmt::Error>;
+
+    fn try_from(f: fmt::Arguments<'a>) -> Result<Self, Self::Error> {
+        use fmt::Write;
+        let mut v = Self::new();
+        v.write_fmt(f).map_err(|e| CapacityError::new(e))?;
+        Ok(v)
```
bluss, line 608: These two lines can be simplified - just remove the Ok(v) and the ?;

c410-f3r:
> Could you add a test for this method?

Sure!

bluss:
Thanks

c410-f3r:
Thank you @bluss. You are awesome!

PR #172, opened 2020-12-17
Use const generics in ArrayVec

### LINKED ISSUE #100: Re-add generic-array integration
Integration with the [generic-array](https://crates.io/crates/generic-array/) crate was [added](https://github.com/bluss/arrayvec/pull/41) but then [reverted](https://github.com/bluss/arrayvec/commit/f65ea92da49f7b8cda6ce5df6f7ec85d484046eb) with the following message:

```
- It can be added back later if needed
- A public dependency implies version coupling, and we can't afford to
  have it for a niche use case.
```

It's not clear to me what is meant by the version coupling problem; can we clarify?

### LINKED ISSUE #125: Support for constant generics
Constant generics is becoming more mature and soon will be in an usable state for complex projects.

bluss:
Use const generics for ArrayVec and ArrayString - now it can support arbitrary capacity/backing array.

Syntax changes like this:

- old: `ArrayVec<[T; CAP]>`
- **new: `ArrayVec<T, CAP>`**
- old: `ArrayString<[u8; CAP]>`
- **new: `ArrayString<CAP>`**

The `ArrayVec<T, const CAP: usize>` for now always uses an `usize` for its length, and so uses more space than the old `ArrayVec<[T; CAP]>`.

Requires Rust 1.51 (March 2021)

In for example Vim one can use the following substitution to update syntax:

    :%s/ArrayVec\(:\?:\?\)<\[\(.*\);\(.*\)\]>/ArrayVec\1<\2,\3>/g

Fixes #125
Closes #100


src/const.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// Return the capacity of the `ArrayVec`.
+    ///
+    /// ```
+    /// use arrayvec::ArrayVec;
+    ///
+    /// let array = ArrayVec::from([1, 2, 3]);
+    /// assert_eq!(array.capacity(), 3);
+    /// ```
+    #[inline(always)]
+    pub fn capacity(&self) -> usize { CAP }
```
mriise, line 91: couldn't this be `const`?

src/const.rs
```diff
@@ -0,0 +1,174 @@
...
+    ///
+    /// ```
+    /// use arrayvec::ArrayVec;
+    ///
+    /// let mut array = ArrayVec::from([1, 2, 3]);
+    /// array.pop();
+    /// assert_eq!(array.len(), 2);
+    /// ```
+    #[inline]
+    pub fn len(&self) -> usize { self.len as usize }
```
mriise, line 68: len is already usize

src/const.rs
bluss: That is not the focus of this PR

bluss:
CI passes on beta and nightly - will pass fully when Rust 1.51 is released this week.

WiSaGaN:
> The ArrayVec<T, const CAP: usize> for now always uses an usize for its length, and so uses more space than the old ArrayVec<[T; CAP]>.

I am wondering if https://github.com/rust-lang/rust/issues/76560 can help with the size regression introduced here in future version of `ArrayVec`?

bluss:
Later PR updated the length to always be u32 (for a smaller size). I can't answer your question, maybe you know? :slightly_smiling_face:

WiSaGaN:
It seems so to me, but I am not terribly familiar with Rust constant generics. Will try push a pull request when it is stable!

PR #174, opened 2021-02-19
Mirror optimization of std::Vec::retain (rust-lang/rust#81126)

niklasf:
Looks like the optimization from rust-lang/rust#81126 applies also to ArrayVec.

### DIFF
--- a/src/arrayvec.rs
+++ b/src/arrayvec.rs
@@ -451,14 +451,14 @@ impl<T, const CAP: usize> ArrayVec<T, CAP> {
         let original_len = self.len();
         unsafe { self.set_len(0) };
 
-        struct BackshiftOnDrop<'a, A: Array> {
-            v: &'a mut ArrayVec<A>,
+        struct BackshiftOnDrop<'a, T, const CAP: usize> {
+            v: &'a mut ArrayVec<T, CAP>,
             processed_len: usize,
             deleted_cnt: usize,
             original_len: usize,
         }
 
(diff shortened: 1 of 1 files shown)


src/arrayvec.rs
```diff
@@ -468,22 +468,58 @@ impl<A: Array> ArrayVec<A> {
...
                 }
             }
         }
-        if del > 0 {
-            self.drain(len - del..);
+
+        let mut g = BackshiftOnDrop { v: self, processed_len: 0, deleted_cnt: 0, original_len };
+
+        while g.processed_len < original_len {
+            let cur = unsafe { &mut *g.v.as_mut_ptr().add(g.processed_len) };
```
bluss, line 506: Now I'm not sure I understand miri correctly but I think you'd want this to be a raw pointer
```suggestion
            let cur = unsafe { g.v.as_mut_ptr().add(g.processed_len) };
```

src/arrayvec.rs
```diff
@@ -468,22 +468,58 @@ impl<A: Array> ArrayVec<A> {
...
+            if !f(cur) {
```
bluss, line 507:
```suggestion
            if !f(&mut *cur) {
```

niklasf:
Ah, indeed. I was confused how this could be fine in std and not here, but the closure in std takes a shared reference.

bluss:
I pushed a fix to the doc build issue in CI, and we can merge this after it runs ok again

bluss (approved):

bluss:
Rebased and updated syntactically for const generics

PR #175, opened 2021-03-23
Panic in .extend() and from_iter on capacity exceeded

### LINKED ISSUE #163: Suggested API change – extend/collect should panic when capacity overflows
I noticed an API inconsistency. `ArrayVec::push` panics if you overflow the capacity, but `ArrayVec::extend` silently drops elements if you overflow the capacity.

I prefer the panicking behavior, since otherwise you run the risk of accidentally losing iterator elements. imo that's not an intuitive failure mode. I'd suggest changing `extend` to panic on overflow, and introduce a fallible `try_extend` (after all, you already have `try_extend_from_slice`), `try_from_iterator`, and probably `try_collect` and a trait to go with it.

This is a breaking change. The migration path would be pretty straightforward:

```rust
// previous version
arrayvec.extend(iterator);

// next version
arrayvec.extend(iterator.take(arrayvec.remaining_capacity());
// -- OR --
let _ignore_result = arrayvec.try_extend(iterator);
```

bluss:
This regresses performance of the .extend(s) benchmark where s is a
slice; to compensate add a private extend_from_slice method that we can
use where possible (clone_from, try_from).

Fixes #163

### DIFF
--- a/src/arrayvec.rs
+++ b/src/arrayvec.rs
@@ -704,7 +704,7 @@ impl<T, const CAP: usize> std::convert::TryFrom<&[T]> for ArrayVec<T, CAP>
             Err(CapacityError::new(()))
         } else {
             let mut array = Self::new();
-            array.extend(slice.iter().cloned());
+            array.extend_from_slice(slice);
             Ok(array)
         }
     }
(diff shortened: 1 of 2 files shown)


bluss:
CI success on beta and nightly (stable will come when Rust 1.51 is released shortly).

PR #177, opened 2021-03-24
Use u32 for the length field in arrayvec

bluss:
Store the length as u32 internally. This is to shrink the size of the
ArrayVec value (when possible, depending on element type).

Inline storage vectors larger than u32::MAX are very unlikely to be
useful - for these cases, prefer using Vec instead.

It's not possible to have the CAP type parameter be of type u32 (missing
features in const evaluation/const generics). We also have to panic at
runtime instead of having a static assertion for capacity, for similar
reasons.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,11 @@ Recent Changes (arrayvec)
 
   The New type syntax is `ArrayVec<T, CAP>` where `CAP` is the arrayvec capacity.
   For arraystring the syntax is `ArrayString<CAP>`.
-  Change by @bluss.
+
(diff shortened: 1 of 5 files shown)


bluss:
cc @clarfonthey since it's related to the `size_of::<ArrayVec<T, CAP>>`

clarfonthey:
I think, personally, it's fine to just keep it at `usize` and then change it to something more reasonable once const eval in generics becomes stable.

Most of the time these arrays are going to be stored on the stack anyway, so it seems unlikely it'll make a huge difference.

bluss:
I see. I think it's important to do a good job for arrayvecs inline in other structs too, and I don't really see any downsides to this present change to use u32.

PR #180, opened 2021-03-26
Minor fix to macro definition

Xaeroxe:


bluss:
Thanks. With no hygiene for that, it seems to have worked anyway.

PR #181, opened 2021-03-28
Made ArrayVec::new and ArrayString::new const fns

### LINKED ISSUE #93: Give ArrayVec a const-fn constructor
I'm not sure if this is even possible but it would be great if `ArrayVec::new()` would be a `const fn` and hence could be called in places where only constant expressions are allowed (of course behind some feature-flag since const-fn is still unstable).

The reason I'd like this feature is that I work on embedded systems where it's typical to have all memory allocated statically, i.e. for a large buffer one would have something like
```rust
static mut BUFFER: [u8; 4096] = [0; 4096];
```
This works fine for arrays (and Copy types) but it would be great if I could also use an `ArrayVec` there. Again, not sure if it's even possible since `mem::uninitialized()` seems to not be a const-fn (although it should, I think) but maybe/hopefully I'm wrong. :)

`lazy_static!` will not work AFAIK because on first invocation of the initialization code the value gets constructed on the stack which typically will not be large enough to hold that value, even temporarily.

Thanks for considering this! :)

rodrimati1992:
Followup to this comment: https://github.com/bluss/arrayvec/issues/93#issuecomment-808920790

Fixes #93

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,9 +29,6 @@ jobs:
           - rust: nightly
             features: serde
             experimental: false
-          - rust: nightly
-            features: serde unstable-const-fn
-            experimental: true
 
     steps:
       - uses: actions/checkout@v2
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -14,6 +14,7 @@ use std::str::Utf8Error;
 use crate::CapacityError;
 use crate::LenUint;
 use crate::char::encode_utf8;
+use crate::utils::MakeMaybeUninit;
 
 #[cfg(feature="serde")]
 use serde::{Serialize, Deserialize, Serializer, Deserializer};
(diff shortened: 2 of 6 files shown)


bluss:
Nice, thanks! Instead of commenting, I just pushed a tweak to the formatting and wording for the feature flag. Thanks for remembering to update that.

bluss:
I considered this, as a follow-up to your solution:

```rust
[(); u32::MAX as usize /*ArrayVec: largest supported capacity is u32::MAX*/][$cap]
```

It also seems to work, but here I'd be worried about compiler quirks that would give problems for such a big array (even if it has 0 size in memory), for example in debuggers.

rodrimati1992:
I have same concern with instantiating long zero-sized arrays everywhere.

bluss (approved):

bluss:
Thanks. It's quite valuable for us to get off nightly features, even if the usage was very limited here.

PR #186, opened 2021-05-24
Use `truncate` in `clone_from`

a1phyr:
This is more efficient because it calls `drop_in_place` on the tail instead on `pop`ing each element separately.


bluss:
Thanks!

PR #187, opened 2021-05-25
take and into_inner_unchecked functions

conradludgate:
Adds two new functions

1. `into_inner_unchecked`, unsafe and unchecked variant of `into_inner`
2. `take`, which returns the ArrayVec while only borrowing, replacing the original with a new empty ArrayVec

### DIFF
--- a/src/arrayvec.rs
+++ b/src/arrayvec.rs
@@ -645,6 +645,7 @@ impl<T, const CAP: usize> ArrayVec<T, CAP> {
     /// Safety:
     /// This operation is safe if and only if length equals capacity.
     pub unsafe fn into_inner_unchecked(self) -> [T; CAP] {
+        debug_assert_eq!(self.len(), self.capacity());
         let self_ = ManuallyDrop::new(self);
         let array = ptr::read(self_.as_ptr() as *const [T; CAP]);
         array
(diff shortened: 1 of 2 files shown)


bluss:
IMO .take() should be reserved for something that takes self, i.e ".take() -> ArrayVec" just like how Option::take works. That way, it can be `take_array` here instead.

conradludgate:
> Would it make sense to write ".take() -> ArrayVec" here and let the user compose it themselves with .into_inner()?

Not a terrible idea. I'll implement that

bluss:
Thanks, this makes sense to include once docs are written and PR description is updated. :slightly_smiling_face:

conradludgate:
I'll update the PR with docs and tests later today then :)

src/arrayvec.rs
```diff
@@ -640,12 +640,25 @@ impl<T, const CAP: usize> ArrayVec<T, CAP> {
         }
     }
 
+    /// Return the inner fixed size array.
+    ///
+    /// Safety:
+    /// This operation is safe if and only if length equals capacity.
     pub unsafe fn into_inner_unchecked(self) -> [T; CAP] {
         let self_ = ManuallyDrop::new(self);
```
bluss, line 648: Let's put in a debug assertion in this one by the way, to make it debug-checked. I or you can add it before release.

bluss (approved):

src/arrayvec.rs
conradludgate: Good point, will get that added now

bluss:
Thanks!

PR #191, opened 2021-07-15
Implement `Index` and `IndexMut` traits for `ArrayVec`

### LINKED ISSUE #190: Feature request: Implement `Index` and `IndexMut`
For code which is generic over the `Index` or `IndexMut` traits, it would be helpful if `ArrayVec` would implement the `Index` and `IndexMut` traits (in `core::ops`). I understand and like that you prefer `ArrayVec` methods not to panic but rather return a `Result` or `Option`, and I see I can use the `get` and `get_mut` functions and then `unwrap` myself, but for code generic over these traits this does not help (much).

Precedence for implementing `Index<I>` and `IndexMut<I>` where `I: SliceIndex<[T]>` (even though the methods may panic):
* (primitive) arrays
* `std::vec::Vec`
* `smallvec::SmallVec` (from the popular `smallvec` crate)
* `tinyvec::ArrayVec` (from the `tinyvec` crate)

niluxv:
I didn't have time to add tests to it yet, but this should implement `Index<I>` and `IndexMut<I>` where `I: SliceIndex<[T]>`.

Fixes #190

### DIFF
--- a/src/arrayvec.rs
+++ b/src/arrayvec.rs
@@ -1002,37 +1002,21 @@ impl<T, const CAP: usize> Extend<T> for ArrayVec<T, CAP> {
     }
 }
 
-/// Perform the indexing (`container[index]`) operation.
-///
-/// # Panics
-/// Panics if index is out of bounds.
 impl<I, T, const CAP: usize> Index<I> for ArrayVec<T, CAP>
 where
     [T]: Index<I>,
 {
     type Output = <[T] as Index<I>>::Output;
 
-    /// Performs the indexing (`container[index]`) operation.
-    ///
-    /// # Panics
-    /// Panics if index is out of bounds.
     fn index(&self, index: I) -> &Self::Output {
         &(**self)[index]
     }
 }
 
-/// Perform the mutable indexing (`container[index]`) operation.
-///
-/// # Panics
-/// Panics if index is out of bounds.
 impl<I, T, const CAP: usize> IndexMut<I> for ArrayVec<T, CAP>
 where
     [T]: IndexMut<I>,
 {
-    /// Perform the mutable indexing (`container[index]`) operation.
-    ///
-    /// # Panics
-    /// Panics if index is out of bounds.
     fn index_mut(&mut self, index: I) -> &mut Self::Output {
         &mut (**self)[index]
     }
--- a/src/arrayvec.rs
+++ b/src/arrayvec.rs
@@ -2,7 +2,7 @@
 use std::cmp;
 use std::iter;
 use std::mem;
-use std::ops::{Bound, Deref, DerefMut, RangeBounds};
+use std::ops::{Bound, Deref, DerefMut, Index, IndexMut, RangeBounds};
 use std::ptr;
 use std::slice;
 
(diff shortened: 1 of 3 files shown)


niluxv:
Instead of this implementation (`self.get[_mut](index).unwrap()`) we could also use
```rust
self.deref().index[_mut](index)
```
Maybe that gives better error messages in some situations? I don't know...

bluss:
Looks to be in the right direction, that's nice - I'll be a bit unreachable during the summer, but back in a bit

niluxv:
No problem.

I changed the implementation as described above because it indeed gives much better error messages on out of bounds errors, and added some tests.

src/arrayvec.rs
```diff
@@ -1002,6 +1002,42 @@ impl<T, const CAP: usize> Extend<T> for ArrayVec<T, CAP> {
...
+    I: slice::SliceIndex<[T]>,
+{
+    type Output = <I as slice::SliceIndex<[T]>>::Output;
+
+    /// Performs the indexing (`container[index]`) operation.
+    ///
+    /// # Panics
+    /// Panics if index is out of bounds.
+    fn index(&self, index: I) -> &Self::Output {
+        self.deref().index(index)
```
bluss, line 1020: Now I guess this is a matter of taste, but IMO we don't call operator methods explicitly, we just use the relevant operator. The operator for deref is `*` and index is, you know, etc.. :slightly_smiling_face: 

```suggestion
        (**self)[index]
```

src/arrayvec.rs
```diff
@@ -1002,6 +1002,42 @@ impl<T, const CAP: usize> Extend<T> for ArrayVec<T, CAP> {
...
+impl<I, T, const CAP: usize> IndexMut<I> for ArrayVec<T, CAP>
+where
+    I: slice::SliceIndex<[T]>,
+{
+    /// Perform the mutable indexing (`container[index]`) operation.
+    ///
+    /// # Panics
+    /// Panics if index is out of bounds.
+    fn index_mut(&mut self, index: I) -> &mut Self::Output {
+        self.deref_mut().index_mut(index)
```
bluss, line 1037:
```suggestion
        (**self)[index]
```

bluss (changes requested):
Thanks, yes not regressing the panic messages is certainly a requirement. One could also debate if `I: SliceIndex` or if `[T]: Index<I>` is the trait bound we should use (the latter being yet more general), but it seems fine as it is now.

We need the corresponding change for ArrayString too, but it can be a different PR.

src/arrayvec.rs
niluxv: Usually I also prefer the syntactic sugar, but when implementing `index` I always use `index` explicitly. Mainly a matter of preference. Changed it now. (The generated mir seems nearly identical. The new version uses only 1 additional local and assignment, so I don't expect a difference in runtime performance.)

niluxv:
So I changed the trait bounds as you suggested and just now added implementations to `ArrayString` analogous to those for `ArrayVec`.
I also removed the documentation I initially added to the `Index`/`IndexMut` implementation since the comment on `panic`ing needn't be true for all instantiations of the generic parameter.

bluss (approved):

mikwielgus:
Hello, may I ask what is the status of this PR? I see that this was *approved* but not merged, is there anything stalling this beyond merge conflict resolution? It would be convenient for me to have `Index` implemented for `ArrayVec`.

PR #192, opened 2021-07-16
Add Copy to ArrayVec enabled by a feature

pYtoner:
Completes https://github.com/bluss/arrayvec/issues/32.

Unfortunately, because trait bounds other than Sized are not allowed for const functions the `new_const` function is not available if the copy feature is enabled.

The whole idea behind the whole thing is that there is no way to construct an `ArrayVec<T>` that isn't `T: Copy` if the `copy` feature is enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ harness = false
 [features]
 default = ["std"]
 std = []
+copy = []
 
 [profile.bench]
 debug = true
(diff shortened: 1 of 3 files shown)


bluss:
Hi, this is a non-additive feature and the general guidelines for libraries are that features have to be additive, so this can not be included in this way, unfortunately. There should be a separate ArrayVecCopy type for this use case.

HughHoyland:
That's regrettable. I would totally expect a stack-allocateable `ArrayVec` to be `Copy`. That's the main use case of fixed-size stack-allocateable `Vec` for me.

PR #193, opened 2021-07-17
added ArrayVecCopy

pYtoner:
Made a copy of ArrayVec (ArrayVecCopy) that can only hold Copy items.

This means copying all of its relevant functions and relevant tests.

Closes https://github.com/bluss/arrayvec/issues/32

### DIFF
--- a/src/arrayvec_copy.rs
+++ b/src/arrayvec_copy.rs
@@ -44,6 +44,8 @@ pub struct ArrayVecCopy<T: Copy, const CAP: usize> {
     len: LenUint,
 }
 
+impl<T: Copy, const CAP: usize> Copy for ArrayVecCopy<T, CAP> {}
+
 macro_rules! panic_oob {
     ($method_name:expr, $index:expr, $len:expr) => {
         panic!(concat!("ArrayVecCopy::", $method_name, ": index {} is out of bounds in vector of length {}"),
--- a/src/arrayvec_copy.rs
+++ b/src/arrayvec_copy.rs
@@ -1,0 +1,1216 @@
+
+use std::cmp;
+use std::iter;
+use std::mem;
+use std::ops::{Bound, Deref, DerefMut, RangeBounds};
+use std::ptr;
+use std::slice;
+
+// extra traits
+use std::borrow::{Borrow, BorrowMut};
+use std::hash::{Hash, Hasher};
+use std::fmt;
+
+#[cfg(feature="std")]
+use std::io;
+
+use std::mem::ManuallyDrop;
+use std::mem::MaybeUninit;
+
+#[cfg(feature="serde")]
+use serde::{Serialize, Deserialize, Serializer, Deserializer};
+
+use crate::LenUint;
+use crate::errors::CapacityError;
+use crate::arrayvec_impl::ArrayVecImpl;
+
+/// A vector with a fixed capacity which implements `Copy` and its elemenents are constrained to also be `Copy`.
+///
+/// The `ArrayVecCopy` is a vector backed by a fixed size array. It keeps track of
+/// the number of initialized elements. The `ArrayVecCopy<T, CAP>` is parameterized
+/// by `T` for the element type and `CAP` for the maximum capacity.
+///
+/// `CAP` is of type `usize` but is range limited to `u32::MAX`; attempting to create larger
+/// arrayvecs with larger capacity will panic.
+///
+/// The vector is a contiguous value (storing the elements inline) that you can store directly on
+/// the stack if needed.
+///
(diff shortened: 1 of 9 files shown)


bluss:
Looks to be in the right direction, that's nice - I'll be a bit unreachable during the summer, but back in a bit

bluss:
Ok great, ArrayVecImpl has made it possible to share some core code. Any other ideas for how to share more? This is quite a lot of code to duplicate, even of probably half of it is necessary.

src/arrayvec_copy.rs
```diff
@@ -0,0 +1,1218 @@
...
+                self.v.set_len(new_len);
+                Some(ptr::read(self.v.get_unchecked_ptr(new_len)))
+            }
+        }
+    }
+}
+
+impl<T: Copy, const CAP: usize> ExactSizeIterator for IntoIter<T, CAP> { }
+
+impl<T: Copy, const CAP: usize> Drop for IntoIter<T, CAP> {
```
bluss, line 846: This iterator doesn't need to implement Drop - because `T` is Copy.

src/arrayvec_copy.rs
```diff
@@ -0,0 +1,1218 @@
...
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_list()
+            .entries(&self.v[self.index..])
+            .finish()
+    }
+}
+
+/// A draining iterator for `ArrayVecCopy`.
+pub struct Drain<'a, T: Copy + 'a, const CAP: usize> {
```
bluss, line 881: Drain could be a candidate for sharing between the two implementations, even if it requires a bit of refactoring.

src/lib.rs
```diff
@@ -57,3 +59,6 @@ pub use crate::array_string::ArrayString;
 pub use crate::errors::CapacityError;
 
 pub use crate::arrayvec::{ArrayVec, IntoIter, Drain};
+
+#[cfg(feature="copy")]
+pub use crate::arrayvec_copy::ArrayVecCopy;
```
bluss, line 64: There are other public items inside  `arrayvec_copy` that we need to make reachable for the users - for example its iterator types. It's likely it needs to be in its own module, what do you think?

src/arrayvec_copy.rs
```diff
@@ -0,0 +1,1218 @@
...
+impl<T: Copy, const CAP: usize> ArrayVecCopy<T, CAP> {
+    /// Extend the arrayvec from the iterable.
+    ///
+    /// ## Safety
+    ///
+    /// Unsafe because if CHECK is false, the length of the input is not checked.
+    /// The caller must ensure the length of the input fits in the capacity.
+    pub(crate) unsafe fn extend_from_iter<I, const CHECK: bool>(&mut self, iterable: I)
+        where I: IntoIterator<Item = T>
+    {
```
bluss, line 992: Maybe this method can move to the ArrayVecImpl trait?

pYtoner:
Nice to see you are back. I will do the changes to make it better. This was more of a quick workaround for those needing this badly...

src/lib.rs
pYtoner: Yes, its own module would make using it easier.

pYtoner:
Doing this with a macro would be the easiest way to remove loads of code duplication. Is that something you think would be acceptable for this crate? (Some people don't like macros 😅)

bluss:
Hi. Please use rebase and not merge to update the PR. Sorry I've been away from github a lot. Macros are ok, I like them, also wrote some eyesore ones myself. Now it's hard to judge in every case, but let's say that we want method bodies to be easy to read. I wouldn't put a whole module's worth of code through 1 macro, maybe it can be per method (or group of smaller trait impls).

pYtoner:
I am not sure what I did to my repo I am going to start over.

bluss:
Don't worry. I think you could have force-pushed to your PR branch if you wanted. Doesn't matter so much now.

jacobsvante:
I would love to see this functionality in arrayvec. @pYtoner what was left to finish with this?

pYtoner:
I had other stuff to do so I forgot about this @jacobsvante. :/ But I'll look into whats changed in the crate since then. Maybe this is a lot easier to implement now.

pYtoner:
So the crate is pretty much the same as when I tried to do this PR. The problem was that I had skill issues with git..

Not sure if this crate is still maintained though but I'll fix my PR.

jacobsvante:
@bluss Are you still maintaining this repo?

pYtoner:
@jacobsvante Alright I opened a new one: https://github.com/bluss/arrayvec/pull/261

jacobsvante:
Thanks @pYtoner 😀

PR #194, opened 2021-08-01
Upgrade length/capacity-related functions to const-fn

bhgomes:
The following are upgraded:
- `Array::len`
- `Array::is_empty`
- `Array::capacity`
- `Array::is_full`
- `Array::remaining_capacity` (added one to `ArrayString` because it didn't exist)
- `CapacityError::new`

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -152,6 +152,19 @@ impl<const CAP: usize> ArrayString<CAP>
     /// ```
     pub fn is_full(&self) -> bool { self.len() == self.capacity() }
 
+    /// Returns the capacity left in the `ArrayString`.
+    ///
+    /// ```
+    /// use arrayvec::ArrayString;
+    ///
+    /// let mut string = ArrayString::<3>::from("abc").unwrap();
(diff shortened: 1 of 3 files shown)


c410-f3r:
`Arrayvec::new` was `const` but then was renamed to `Arrayvec::new_const` because of bad codegen (now `Arrayvec::new` is just a regular method).
So, I guess an assembly demonstration of before and after will improve the changes of merging. Or maybe such thing won't be necessary, all depends on @bluss :)
IMHO, all methods should be `const` when possible and bad codegen should be a thing fixed by MIR or LLVM.

bhgomes:
Not sure if this is an issue in this case since these are just accessing struct fields or performing integer arithmetic, but I could be mistaken.

bluss:
Thanks, yes certainly it would be expected that these methods *can't* run into the same problem that the constructor did, in particular in that case we changed the body/implementation to make it possible to make the functions const, and we don't do that here.

bluss (approved):

PR #197, opened 2021-08-13
Add `zero_filled` constructor

### LINKED ISSUE #196: [ArrayString] Infallible constructor to fill inner array
Some libraries like `hex` or `bs58` expect a pre-filled buffer to operate on but `ArrayString` does not have an infallible constructor to create a fully filled inner array with "whatever".

For example, the following made use of `from_byte_string` to make everything work as expected:

```rust
let mut buffer = ArrayString::from_byte_string(&[b' '; SOME_LEN]).unwrap();
let len = bs58::encode(some_data).into(&mut buffer[..]).unwrap_or_default();
buffer.truncate(len);
```

So do you mind me providing a PR to create something like `ArrayString::new_filled()`?

c410-f3r:
Fixes #196

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -129,7 +129,7 @@ impl<const CAP: usize> ArrayString<CAP>
         Ok(vec)
     }
 
-    /// Creates a new `ArrayString` instance fully filled with ASCII NULL characters (`\0`). Useful
+    /// Create a new `ArrayString` value fully filled with ASCII NULL characters (`\0`). Useful
     /// to be used as a buffer to collect external data or as a buffer for intermediate processing.
     ///
     /// ```
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -129,6 +129,28 @@ impl<const CAP: usize> ArrayString<CAP>
         Ok(vec)
     }
 
+    /// Creates a new `ArrayString` instance fully filled with ASCII NULL characters (`\0`). Useful
+    /// to be used as a buffer to collect external data or as a buffer for intermediate processing.
+    ///
+    /// ```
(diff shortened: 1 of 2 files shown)


src/array_string.rs
```diff
@@ -129,6 +129,28 @@ impl<const CAP: usize> ArrayString<CAP>
...
+        let mut vec = Self::new();
+        let data = [b'\0'; CAP];
+        // SAFETY: The array is fully filled with valid UTF-8
+        unsafe {
+            let ptr = data.as_ptr() as *const [MaybeUninit<u8>; CAP];
+            ptr.copy_to_nonoverlapping(&mut vec.xs as *mut _, 1);
+            vec.set_len(CAP);
+        }
+        vec
+    }
```
bluss, line 152: I think we can and should copy the code in `fn new` verbatim, but instead of uninit, we use `MaybeUninit::zeroed` and len: CAP and it should be good

c410-f3r: Done!

bluss (approved):

src/array_string.rs
```diff
@@ -129,6 +129,28 @@ impl<const CAP: usize> ArrayString<CAP>
         Ok(vec)
     }
 
+    /// Creates a new `ArrayString` instance fully filled with ASCII NULL characters (`\0`). Useful
```
bluss, line 132:
```suggestion
    /// Create a new `ArrayString` value fully filled with ASCII NULL characters (`\0`). Useful
```

bluss:
Thanks!

c410-f3r:
Thank you!

PR #205, opened 2021-11-07
Const constructors from str, from slice for ArrayString, ArrayVec

### LINKED ISSUE #204: Const constructor for strings via a macro?
Wonder if it's possible in theory (scratched my head at it for a bit but couldn't quite figure all the details), to have something like:

```rust
const S: ArrayString<10> = array_string!("foo");
```

This would be super cool as it would allow pre-initialized array-strings in const contexts which is currently not doable.

// - Without macros, as of this moment, don't think it's possible
// - It's certainly doable via a proc-macro, but may require adding an additional 'raw' const constructor to ArrayString

bluss:
- Add from_str_const constructor for ArrayString

- Add from_slice_const constructor for ArrayVec

We can't make the ArrayVec constructors generic, not even where T: Copy,
so they are just templated out with macros for the integer types, quite limited.

Closes #204


bluss:
Depends on Rust 1.58 for panicking in const constructors, so it's too early for this. We need some kind of error handling and it seems this is the only one available.

aldanor:
Neat!

About compile-time const, for pre-1.57, you can use a hack like this btw:

```rust
pub(crate) struct Assert<const L: usize, const R: usize>;

impl<const L: usize, const R: usize> Assert<L, R> {
    pub const LENGTH_LT_CAPACITY: usize = R - L - 1;
}

macro_rules! assert_length_lt_capacity_const {
    ($len:expr, $cap:expr) => {
        #[allow(path_statements)]
        { Assert::<$len, $cap>::LENGTH_LT_CAPACITY; }
    }
}

fn main() {
    assert_length_lt_capacity_const!(2, 3);
    // assert_length_lt_capacity_const!(3, 3); // compile-time error
}
```

bluss:
I see. That's similar to what I already added for it to work on stable, but maybe your version has better error messages

aldanor:
It's not perfect, but quite readable. Perhaps could be improved by something like

```rust
impl<const LEN: usize, const CAP: usize> Assert<LEN, CAP> {
    pub const OK: usize = CAP - LEN - 1; /* insufficient array capacity */
}
```

This yields:

```
error: any use of this value will cause an error
 --> src/main.rs:4:27
  |
4 |     pub const OK: usize = CAP - LEN - 1; /* insufficient array capacity */
  |     ----------------------^^^^^^^^^^^^^-
  |                           |
  |                           attempt to compute `0_usize - 1_usize`, which would overflow
```

bluss:
It seems like the length - coming from `s.len()`  can't be used there in current Rust, unfortunately.

aldanor:
> It seems like the length - coming from `s.len()` can't be used there in current Rust, unfortunately.

Ah, yea, indeed, it would only work with "true consts"...

So what's the plan then, wait till MSRV >= 1.57 bump for the crate eventually?

bluss:
It won't work there either (I'm developing on nightly locally, so that's the first behaviour I see). I'm not sure what the missing feature would be called.

aldanor:
Sorry for bringing the old issue back from the dead :) Wonder if this would be feasible, given we can panic in constants?

Const constructors are still a big missing piece...

utkarshgupta137:
Bump

jon-zu:
Is there a reason to use slices instead o an Array `[T; M]` with an assertion that `M <= CAP` (assert works fine now in const fns). I've tried this, and It works just fine, the only downside is Copy is still required, and we have to pass the Array as reference else the compiler compilers that It can't determine the destructor.

I would argue this const constructor for `ArrayString` and one for Arrays covers almost all cases.

GummyGun:
Is there any reason this is not currently in yet. 
Is the crate dead/stalled?

ghost:
Is this still Work in Progress? Or are we just waiting for a feature to stabilize?

PR #207, opened 2021-11-10
ArrayVecCopy

pYtoner:
A new version of `ArrayVecCopy` which uses macros for removing code duplication.

I am not sure how to get the macros to replace the tokens in the docs.

But anyway I am not quite sure about doing it like this. It becomes quite messy. This is not done yet so I am just checking in to see if this is the way we really want to go about doing this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "arrayvec"
-version = "0.7.2"
+version = "0.7.3"
 authors = ["bluss"]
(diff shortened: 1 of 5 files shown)


bluss:
Hm I see. Well, good that you're checking. If you don't believe in this, then we sure can't use it. Using big macros are messy, sure.

I don't have the overview. The biggest mess, the trait bounds and the unreadable macro stuff that needs? It's not looking too bad, maybe for example the serde impl can be shared like this in the future version.

PR #210, opened 2021-12-23
Follow simplification of Vec::retain() (rust-lang/rust#91527)

niklasf:
Follows rust-lang/rust#91527. This is neutral in my go-to benchmarks, but it's probably good to keep the implementation in sync regardless.

Edit: Looks like miri does not like this, whereas it doesn't mind std. Reminds me of https://github.com/bluss/arrayvec/pull/174#pullrequestreview-594795729. Will debug later.

### DIFF
--- a/src/arrayvec.rs
+++ b/src/arrayvec.rs
@@ -493,39 +493,38 @@ impl<T, const CAP: usize> ArrayVec<T, CAP> {
 
         let mut g = BackshiftOnDrop { v: self, processed_len: 0, deleted_cnt: 0, original_len };
 
-        #[inline(always)]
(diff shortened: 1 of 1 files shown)


niklasf:
Fixed by dereferencing the pointer only for `f(unsafe { &mut *cur })`. To be honest, I don't understand how std gets away with not doing it like that.

AngelicosPhosphoros:
@bluss Can we merge that?
It would fix complains of miri about retain.
https://github.com/bluss/arrayvec/issues/221

AngelicosPhosphoros:
@niklasf 
At this moment miri also don't like `ArrayVec::drain` and `ArrayVec::retain`.

niklasf:
Looks like the stacked borrow violations have been addressed in a different way.

PR #214, opened 2022-02-13
pop_unchecked

vorot93:


tbu-:
I know this is 2½ years late.

What is the use case for this function?

PR #217, opened 2022-04-07
Implement AsMut<str> for ArrayString

c410-f3r:
`ArraySting` already implements `DerefMut` so `AsMut` is like a missing piece

It is worth noting that `String` also implements `AsMut`


c410-f3r:
Closing due to the lack of feedback.

PR #220, opened 2022-06-23
Shrink unsafe blocks

peamaeq:
In these function you use the unsafe keyword for some safe expressions. 

We need to mark unsafe operations more precisely using unsafe keyword. Keeping unsafe blocks small can bring many benefits. For example, when mistakes happen, we can locate any errors related to memory safety  within an unsafe block. This is the balance between Safe and Unsafe Rust. The separation is designed to make using Safe Rust as ergonomic as possible, but requires extra effort and care when writing Unsafe Rust. 

Hope this PR can help you.
Best regards.
**References**
https://doc.rust-lang.org/nomicon/safe-unsafe-meaning.html 
https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -435,8 +435,8 @@ impl<const CAP: usize> DerefMut for ArrayString<CAP>
 {
     #[inline]
     fn deref_mut(&mut self) -> &mut str {
+        let len = self.len();
         unsafe {
-            let len = self.len();
             let sl = slice::from_raw_parts_mut(self.as_mut_ptr(), len);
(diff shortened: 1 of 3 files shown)


AngelicosPhosphoros:
@bluss Can we merge that?
It would fix complains of miri about retain.
https://github.com/bluss/arrayvec/issues/221

bluss:
Thanks and my thoughts echo what @tbu- said. While the PR's changes are correct and harmless, this is not a logic for updating unsafe code blocks that I believe in.

PR #222, opened 2022-07-23
Fix Stacked Borrows violations in crate

AngelicosPhosphoros:
It mainly caused by borrowing whole internal array in `as_mut_ptr` and `as_ptr`. Stdlib `Vec` doesn't suffer from this because it keeps pointer to data internally so it doesn't reborrow slice when getting pointers.

This changes required complete rewrite of `Drain` struct and little adjustment of `retain`.
Also, there is little changes in handling ZSTs and in ArrayString.

Also, I run `cargo fmt` on code and cherry-picked commit from #215.

Probably better to review one commit at time.


AngelicosPhosphoros:
Rebased and added some extra tests.

bluss:
Thanks. Are these violations that miri does not check? I.e does this do something beyond fixing errors that the miri CI reports?

bluss:
I have merged PR #226 which is on the same topic, but is easier to review and more in line with what I expect. There might be nuances I don't know about though.

My main crime is not being an active maintainer (and maybe having no love for rustfmt), and I'm sorry for that. I haven't had time for Rust in a long while unfortunately.

PR #226, opened 2022-09-06
Fix stacked borrows violations

clubby789:
Use offsets from a single pointer instead of multiple pointers into an array.
Also removed an integer-pointer cast which can cause issues with Miri


bluss:
Thank you, rebased and retested in CI and it looks good

bluss:
failures reported in master here: https://github.com/bluss/arrayvec/actions/runs/3450267864/jobs/5758757231  and they are fixed by this change

PR #230, opened 2022-11-09
Add missing `BorrowMut` implementation for `ArrayString`

msrd0:
There is already a `BorrowMut` implementation for `ArrayVec` and in std, there is also a `BorrowMut` implementation for `String`.


msrd0:
The miri CI failure does not look like it was caused by this PR.

bluss:
Thanks

PR #233, opened 2023-01-15
Add convenience constructor for null-terminated byte literals

kaidokert:
This adds a small convenience function for making ArrayStrings out of null-terminated C-style byte buffers, by truncating the result to first null-terminator, if present.

Of course this a pretty trivial utility to keep as a free-standing function in code that needs to deal with C input, so maybe not a great fit here.

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -129,6 +129,23 @@ impl<const CAP: usize> ArrayString<CAP>
         Ok(vec)
     }
 
+    /// Create a new `ArrayString` from a byte string literal, that can be null-terminated.
+    ///
+    /// **Errors** if the byte string literal is not valid UTF-8.
+    /// ```
+    /// use arrayvec::ArrayString;
+    ///
+    /// let string = ArrayString::from_c_byte_string(b"hello\0world").unwrap();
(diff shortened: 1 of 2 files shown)


Callum-A:
@bluss anyway we could look to get this merged?

bluss:
Hm it kind of needs a better name or description, doesn't it? Nothing from the current name makes it clear what it does (that it will truncate at the null terminator).

tbu-:
> anyway we could look to get this merged?

You can create this function in your own crate, too, by creating an extension trait. You don't have to wait for this (or similar PRs).

Callum-A:
> > anyway we could look to get this merged?
> 
> You can create this function in your own crate, too, by creating an extension trait. You don't have to wait for this (or similar PRs).

Yep have done so for now, little wrapper struct does this on our side. Would be a nice one to get merged.

Callum-A:
> Hm it kind of needs a better name or description, doesn't it? Nothing from the current name makes it clear what it does (that it will truncate at the null terminator).

Fair point, null terminated could be a bit of a misnomer here as it will break at the first null and truncates. Either way this doesn't have to be urgently looked at as we easily implemented this.

PR #235, opened 2023-04-04
Add ArrayVec::leak

luksan:
Hello,
This pull request adds leak() to ArrayVec, which can be useful if you want to hand out a reference to the contained data, and at the same time want to be sure that the vec is empty when the reference is dropped. Especially useful for !Drop types.

### DIFF
--- a/src/arrayvec.rs
+++ b/src/arrayvec.rs
@@ -579,6 +579,19 @@ impl<T, const CAP: usize> ArrayVec<T, CAP> {
         Ok(())
     }
 
+    /// Leak the contents of the vector and return a reference to them as a mutable slice.
+    ///
+    /// This sets len() to zero. The elements will not be dropped. The backing memory will
+    /// not be lost however, as the ArrayVec can be used again as soon as the returned slice
+    /// is dropped.
+    pub fn leak(&mut self) -> &mut [T] {
+        let orig_len = self.len();
+        unsafe {
+            self.set_len(0);
+            slice::from_raw_parts_mut(self.as_mut_ptr(), orig_len)
+        }
+    }
+
     /// Create a draining iterator that removes the specified range in the vector
     /// and yields the removed items from start to end. The element range is
     /// removed even if the iterator is not consumed until the end.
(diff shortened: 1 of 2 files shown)


bluss:
Why do you want to leak the values?

luksan:
My use case is a reusable buffer. The bytes are pushed into the buffer, and when the content is finalized it can be handed out as a &[u8]. The borrow checker makes sure that the buffer isn't re-used as long as a reference to the contents are live, and the old content will never be seen after the &[u8] ref is dropped. "leak()" isn't the best of names for the operation, as the memory isn't actually leaked, but it aligns with std :: Vec.

bluss:
Names along the lines of take are also interesting then, but they don't convey the leaking (when it applies).

You mentioned that Vec has Vec::leak, and that should be in the first PR description - that's usually a good enough argument for any new arrayvec method. Then it's easy for us to just adapt the same idea.

src/arrayvec.rs
```diff
@@ -579,6 +579,19 @@ impl<T, const CAP: usize> ArrayVec<T, CAP> {
...
 
+    /// Leak the contents of the vector and return a reference to them in a mutable slice.
+    ///
+    /// This sets len() to zero. The elements will not be dropped. The backing memory will
+    /// not be lost however, as the ArrayVec can be used again as soon as the returned slice
+    /// is dropped.
+    pub fn leak(&mut self) -> &mut [T] {
+        unsafe {
+            let s = slice::from_raw_parts_mut(self.as_mut_ptr(), self.len());
+            self.set_len(0);
```
bluss, line 590: I would be more comfortable to swap the order of these two methods actually! 

A. We create a borrow of the whole vec as a &mut [T]
B. We modify self (set_len)

This is better in the order of B, then A, so that it's clear we don't invalidate the borrow. Miri apparently didn't react on this though?

tbu-:
> no github actions run, don't immediately see why

You can enable action runs for first-time contributors in the settings.

bluss:
ok, changed to a more permissive setting

PR #236, opened 2023-04-10
track_caller for capacity overflow panics

kornelski:
'CapacityError: insufficient capacity' panic points to `~/.cargo/registry/src/github.com-1ecc6299db9ec823/arrayvec-0.7.2/src/arrayvec.rs:176:9` instead of the line of user code that pushed. 

`#[track_caller]` makes this error more relevant to users.

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -201,6 +201,7 @@ impl<const CAP: usize> ArrayString<CAP>
     ///
     /// assert_eq!(&string[..], "ab");
     /// ```
+    #[track_caller]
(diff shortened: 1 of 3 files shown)


bluss:
Thanks, it does make sense. track_caller is supposed to have some overhead, but I can't find it quantified anywhere. I suppose it's not API-defining anyway, so worst case it can always be changed around - just saying that to make it easy for us to include this change.

kornelski:
It has overhead of having an extra function argument, but when used with small generic functions, the calls and their arguments get inlined, so the overhead usually disappears.

PR #237, opened 2023-04-11
Add `push_many_from_slice` method

Ericson2314:
See docs for details.


Ericson2314:
https://github.com/alamgu/alamgu-async-block/pull/5 Here is an example of the method being used.

PR #248, opened 2023-09-29
Add len_u16 and len_u8 features.

WuBingzheng:
Add `len_u16` and `len_u8` features to change the type of `LenUint`.

Using `u8` as `LenUint` saves 3 bytes. This is useful for many really small strings.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,8 @@ harness = false
 [features]
 default = ["std"]
 std = []
+len_u16 = []
+len_u8 = []
 
 [profile.bench]
 debug = true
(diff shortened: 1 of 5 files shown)


GnomedDev:
This should not be a feature, as features should be entirely additive. I would recommend making it a generic, maybe?

WuBingzheng:
I do not know how to make it a generic. Use the `num` trait? 

```
pub struct ArrayVec<T, L, const CAP: usize>
where L: num::Num
{
    xs: [MaybeUninit<T>; CAP],
    len: L,
}
```

PR #250, opened 2023-09-30
Added a map method, similar to `array::map` to the `ArrayVec`

miestrode:
Because an `ArrayVec` fundamentally has a capacity, it would be useful to have a map method, kind of like `array::map`. This can be useful for constructing certain `ArrayVec`s from old ones, as can be ascertained. I've written two tests (a doctest and a regular integration test) for the method, but if more need to be written, or if somehow this PR is irrelevant due to some feature I'm unaware of, please let me know.

Example usage:
```rust
let av = ArrayVec::from([1, 2, 3]).map(|element| element * 2);
assert_eq!(&av[..], &[2, 4, 6]);
```

### DIFF
--- a/src/arrayvec.rs
+++ b/src/arrayvec.rs
@@ -583,6 +583,50 @@ impl<T, const CAP: usize> ArrayVec<T, CAP> {
         Ok(())
     }
 
+    /// Returns an `ArrayVec` by applying function `f` to each element of the passed `ArrayVec`.
+    ///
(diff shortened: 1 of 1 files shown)


miestrode:
For some reason I thought I remembered you couldn't collect into an `ArrayVec`. Probably should've rechecked that before filing the PR. Terribly sorry. Kind of curious if this would be in any way faster, but I think I'll leave that for someone else, as I no longer feel the need for it, at least personally.

PR #255, opened 2023-12-19
Put len before xs for better memory layout

### LINKED ISSUE #254: Better memory layout by moving len before xs
Hi, I noticed that `len` is arranged after `xs`
```rust
pub struct ArrayVec<T, const CAP: usize> {
    // the `len` first elements of the array are initialized
    xs: [MaybeUninit<T>; CAP],
    len: LenUint,
}
```
In C's world, it's a convention to move length before an array, like
```rust
#[repr(C)]
pub struct ArrayVec<T, const CAP: usize> {
    len: LenUint,
    // the `len` first elements of the array are initialized
    xs: [MaybeUninit<T>; CAP]
}
```
This provides a few nice features:
- better locality. First, read the length, then read the element one by one. This should be trivial though
- Easier to implement tail arrays if were to put on the heap 
- https://github.com/antirez/sds Simply Dynamic String uses a similar layout

JakkuSakura:
closes #254

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -31,10 +31,11 @@ use serde::{Serialize, Deserialize, Serializer, Deserializer};
 /// The string is a contiguous value that you can store directly on the stack
 /// if needed.
 #[derive(Copy)]
+#[repr(C)]
 pub struct ArrayString<const CAP: usize> {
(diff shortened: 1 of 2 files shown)


bluss:
Thanks

PR #253, opened 2023-12-07
Optional support for borsh serialization

Fuuzetsu:

PR #259, opened 2024-01-18
Improved support for borsh serialization

honzasp:
Builds on #253 with the following improvements:

- Do not allocate when deserializing `ArrayString`
- Serialize length as `u32`, not as `u64`, to be consistent with serialization of `[T]` and `str`
- Add tests

Closes #253


bluss:
Thanks both of you

PR #261, opened 2024-02-09
Add ArrayVecCopy V2

pYtoner:
New version of closed: https://github.com/bluss/arrayvec/pull/193

Made a copy of ArrayVec (ArrayVecCopy) that can only hold Copy items.

This means copying all of its relevant functions and relevant tests.

Closes https://github.com/bluss/arrayvec/issues/32

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -61,3 +61,10 @@ pub use crate::array_string::ArrayString;
 pub use crate::errors::CapacityError;
 
 pub use crate::arrayvec::{ArrayVec, IntoIter, Drain};
+
+#[cfg(feature = "copy")]
+mod arrayvec_copy;
+#[cfg(feature = "copy")]
+pub mod copy {
+    pub use crate::arrayvec_copy::{ArrayVecCopy, IntoIter, Drain};
+}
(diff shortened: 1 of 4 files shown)


pYtoner:
Unfortunately miri says there is some UB in `copy_tests::test_drain`. Not quite sure how to fix it yet.

src/lib.rs
```diff
@@ -61,3 +61,10 @@ pub use crate::array_string::ArrayString;
 pub use crate::errors::CapacityError;
 
 pub use crate::arrayvec::{ArrayVec, IntoIter, Drain};
+
+#[cfg(feature = "copy")]
+mod arrayvec_copy;
+#[cfg(feature = "copy")]
+pub mod copy {
+    pub use crate::arrayvec_copy::{ArrayVecCopy, IntoIter, Drain};
```
tbu-, line 69: Perhaps only keep the iterator types in here and expose the `ArrayVecCopy` globally.

tbu- (commented):
I wonder how to keep the whole thing maintainable. If possible, the `arrayvec_copy` module should be mostly kept in sync with the `arrayvec` module. Perhaps one could put up a diff plus some search/replace that generates the `arrayvec_copy` module?

PR #262, opened 2024-02-28
Make `ArrayString::as_{mut_}ptr` public.

finnbear:
## Motivation

I currently get the pointer via `impl Deref<Target = str>` but Miri complains (stacked borrows) when I write past the end (in low-level serialization code).

`ArrayVec::as_{mut_}ptr` works for `ArrayVec`, so I just need the same thing for `ArrayString`.

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -414,11 +414,13 @@ impl<const CAP: usize> ArrayString<CAP>
         self
     }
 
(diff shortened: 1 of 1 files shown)


finnbear:
Duplicate of recently-opened https://github.com/bluss/arrayvec/pull/260

PR #263, opened 2024-03-11
Remove borsh from msrv builder

bluss:
Is not supported at msrv anyway (msrv applies to defaults)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       matrix:
         include:
           - rust: 1.51.0 # MSRV
-            features: serde, borsh
+            features: serde
             experimental: false
           - rust: stable
             features:


tbu- (approved):
LGTM.

As the compilation output shows, we'd probably have to commit a `Cargo.lock` to make sure that the crate continues to compile on the MSRV.

bluss:
I see, so serde still fails too.

I think I like specific cargo upgrade --precise  calls - for the msrv builder -- instead of using a lockfile?

tbu-:
> I think I like specific cargo upgrade --precise calls - for the msrv builder -- instead of using a lockfile?

That works, it doesn't pin dependencies of dependencies though. What do we want to test? Using a lockfile has the advantage of pinning everything, we see when arrayvec breaks something.

The Cargo FAQ explicitly mentions this use case for a Cargo.lock, from https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control:
> Having this snapshot of dependencies can also help when projects need to be verified against consistent versions of dependencies, like when
>
> * Verifying a minimum-supported Rust version (MSRV) that is less than the latest version of a dependency supports

bluss:
Sounds good, if you want to. I merged here since it was at least a step in the right direction

PR #266, opened 2024-06-09
Automatically document feature flag requirements on docs.rs by using the feature `doc_auto_cfg`

JSorngard:
Thanks for the nice crate!

Since this is an unsolicited PR, I completely understand if it is not interesting.

This PR uses the idea discussed in [this](https://stackoverflow.com/questions/61417452/how-to-get-a-feature-requirement-tag-in-the-documentation-generated-by-cargo-do/) stackoverflow post to automatically document all feature flag requirements on docs.rs. This way any future feature flag gated code does not have to worry about documenting what features it needs.

The feature requires a nightly compiler, but by conditionally enabling it only on docs.rs the MSRV of the crate is unchanged (docs.rs always uses the nightly compiler).

The result should look like this:
bild  
and this:  
bild

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,7 @@ debug = true
 debug = true
 
 [package.metadata.docs.rs]
-features = ["borsh", "serde", "zeroize"]
+all-features = true
 # Enables the `docsrs` config attribute when running on docs.rs,
 # which always uses the latest nightly compiler. 
 # Idea taken from <https://stackoverflow.com/a/70914430>.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,9 @@ debug = true
 
 [package.metadata.docs.rs]
 features = ["borsh", "serde", "zeroize"]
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -54,7 +54,11 @@ debug = true
 debug = true
 
 [package.metadata.docs.rs]
-features = ["borsh", "serde", "zeroize"]
+all-features = true
+# Enables the `docsrs` config attribute when running on docs.rs,
+# which always uses the latest nightly compiler. 
+# Idea taken from <https://stackoverflow.com/a/70914430>.
+rustdoc-flags = ["--cfg", "docsrs"]
```
bluss, line 61: Nice but probably not needed anymore https://docs.rs/about/builds

bluss:
That looks very good but right now, with the slow speed of development, I don't want to use any unstable features.

JSorngard:
Ah, I hadn't seen that! Very nice.

Makes sense to not merge. Should I delete the PR, or is it something for the future?

PR #269, opened 2024-08-16
Allow concatenation of different array strings

Callum-A:
As a user I would like to concat two `ArrayString` easily. This exposes two methods to do this.

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -241,6 +241,68 @@ impl<const CAP: usize> ArrayString<CAP>
         }
     }
 
+    /// Adds the given `ArrayString` to the end of this `ArrayString`.
+    ///
+    /// ***Panics*** if the backing array is not large enough to fit the additional `ArrayString`.
+    ///
+    /// ```
+    /// use std::str::FromStr;
+    /// use arrayvec::ArrayString;
+    ///
+    /// let mut string = ArrayString::<2>::new();
+    /// let a = ArrayString::<1>::from_str("a").unwrap();
+    /// let b = ArrayString::<1>::from_str("b").unwrap();
+    ///
+    /// string.push_array_string(a);
+    /// string.push_array_string(b);
+    ///
+    /// assert_eq!(&string[..], "ab");
+    /// ```
+    #[track_caller]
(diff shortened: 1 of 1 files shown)


src/array_string.rs
```diff
@@ -241,6 +241,68 @@ impl<const CAP: usize> ArrayString<CAP>
...
+    /// let a = ArrayString::<1>::from_str("a").unwrap();
+    /// let b = ArrayString::<1>::from_str("b").unwrap();
+    ///
+    /// string.push_array_string(a);
+    /// string.push_array_string(b);
+    ///
+    /// assert_eq!(&string[..], "ab");
+    /// ```
+    #[track_caller]
+    pub fn push_array_string<const OTHER_CAP: usize>(&mut self, other: ArrayString<OTHER_CAP>) {
```
bluss, line 262: The user should just use `string.push_str(&a)` here. Is there a reason not to do that?

src/array_string.rs
```diff
@@ -241,6 +241,68 @@ impl<const CAP: usize> ArrayString<CAP>
...
+    /// ```
+    pub fn try_push_array_string<const OTHER_CAP: usize>(
+        &mut self,
+        other: ArrayString<OTHER_CAP>,
+    ) -> Result<(), CapacityError<ArrayString<OTHER_CAP>>> {
+        if other.len() > self.capacity() - self.len() {
+            return Err(CapacityError::new(other));
+        }
+        unsafe {
+            let dst = self.as_mut_ptr().add(self.len());
```
bluss, line 296: (Not relevant unless this will be added) - this is duplicated code and should just reuse `try_push_str`

src/array_string.rs
Callum-A: Ah would this work? I actually didn't know that! I assume this would also avoid heap allocs?

bluss: of course, no heap allocations extra. ArrayString is just like another kind of String - they both have zero cost deref to &str.

Callum-A: Ah perfect this can likely get closed out then!

bluss: Ok, feel free to add an example that shows how this works, if you want to.

bluss:
Closing as the use case is already served by `.push_str(&x)`.

PR #278, opened 2024-10-15
Add try_push_uninit and test

phughk:
Add try_push_uninit , which tries to add an additional element and provides a raw pointer to the already allocated memory.

This is useful if the backing structure is larger than the stack size.

Included is a test that demonstrates the need for this functionality.


src/arrayvec.rs
```diff
@@ -206,6 +206,22 @@ impl<T, const CAP: usize> ArrayVec<T, CAP> {
         ArrayVecImpl::try_push(self, element)
     }
 
+    /// Push a new uninitialised value to the end of the vector and return
+    /// a mutable reference to it.
+    ///
+    /// This is useful if the backed value is large and won't fit on stack.
+    ///
+    /// Capacity error carries the data - since we don't have data then we need a different error type.
```
bluss, line 214: please use `CapacityError<()>`

bluss:
Something like this is needed, but unsure about the right API

tbu-:
Perhaps this could take the same shape as [`Vec::spare_capacity_mut`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.spare_capacity_mut)?

bluss:
@tbu- that sounds good!

bluss:
@tbu-  was quick to add a suggestion in #279 which is a pretty nice interface - and more versatile in some ways - I hope that addresses the need here, what do you think?

phughk:
@tbu- @bluss thank you! Yeah I'm not obsessed about the PR, I added this to a fork and used in my own project and wanted to contribute back. If you have a better API then very happy with the turnaround. Thank you as well for the diligent review; I learnt from it 😊

PR #279, opened 2024-10-16
Add `ArrayVec::spare_capacity_mut`

tbu-:
This mirrors `Vec::spare_capacity_mut`. Description and example are taken from this function, too.

CC #278

### DIFF
--- a/src/arrayvec.rs
+++ b/src/arrayvec.rs
@@ -535,6 +535,41 @@ impl<T, const CAP: usize> ArrayVec<T, CAP> {
         drop(g);
     }
 
+    /// Returns the remaining spare capacity of the vector as a slice of
+    /// `MaybeUninit<T>`.
+    ///
+    /// The returned slice can be used to fill the vector with data (e.g. by
+    /// reading from a file) before marking the data as initialized using the
+    /// [`set_len`] method.
+    ///
+    /// [`set_len`]: ArrayVec::set_len
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use arrayvec::ArrayVec;
+    ///
+    /// // Allocate vector big enough for 10 elements.
+    /// let mut v: ArrayVec<i32, 10> = ArrayVec::new();
+    ///
+    /// // Fill in the first 3 elements.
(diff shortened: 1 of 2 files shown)


src/arrayvec.rs
```diff
@@ -535,6 +535,41 @@ impl<T, const CAP: usize> ArrayVec<T, CAP> {
...
+    /// use arrayvec::ArrayVec;
+    ///
+    /// // Allocate vector big enough for 10 elements.
+    /// let mut v: ArrayVec<i32, 10> = ArrayVec::new();
+    ///
+    /// // Fill in the first 3 elements.
+    /// let uninit = v.spare_capacity_mut();
+    /// uninit[0].write(0);
+    /// uninit[1].write(1);
+    /// uninit[2].write(2);
```
tbu-, line 559: The test failure is about [`MaybeUninit::write`](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.write-1) only being stable 1.55. What's the preferred course of action?
1. Use a method that was stable in 1.55.
2. Ignore this doctest when running with 1.55.

bluss: just ignore testing the example is to be preferred

src/arrayvec.rs
```diff
@@ -535,6 +535,41 @@ impl<T, const CAP: usize> ArrayVec<T, CAP> {
...
     }
 
+    /// Returns the remaining spare capacity of the vector as a slice of
+    /// `MaybeUninit<T>`.
+    ///
+    /// The returned slice can be used to fill the vector with data (e.g. by
+    /// reading from a file) before marking the data as initialized using the
+    /// [`set_len`] method.
+    ///
+    /// [`set_len`]: Vec::set_len
```
bluss, line 545: link still points to vec here

bluss:
This is good, nice interface and replicating Vec should almost always be welcome in ArrayVec

src/arrayvec.rs
tbu-: Fixed.

bluss:
Ok thank you and let's go

PR #284, opened 2024-12-20
Return a proper error for `into_inner`

patrickariel:
This PR adds a new error type representing an under-filled state for `ArrayVec::into_inner`. Previously, it would return `self` which is rather unconventional and prevents ergonomic error propagation (e.g. with `?`).

### DIFF
--- a/src/arrayvec.rs
+++ b/src/arrayvec.rs
@@ -21,7 +21,7 @@ use std::mem::MaybeUninit;
 use serde::{Serialize, Deserialize, Serializer, Deserializer};
 
 use crate::LenUint;
-use crate::errors::CapacityError;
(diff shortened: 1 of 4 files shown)


tbu-:
This change is a breaking change and would need a new major version.

After the change, one could no longer recover the original array.

Workaround for your use case: You can use `.map_err(|_| YourError::Underfilled)?` to use it with the question mark operator.

patrickariel:
Thanks for the feedback, I changed it so that the error stores the original vec, allowing the user to recover it using `take_vec`.

> Workaround for your use case: You can use `.map_err(|_| YourError::Underfilled)?` to use it with the question mark operator.

This is what I'm currently doing, but it'd more convenient to have a canonical error type :slightly_smiling_face:

PR #289, opened 2025-01-30
Add from iterator impl

Pzixel:
This is the best effort implementation that allows pushing iterator into `ArrayString`. There is no `TryFromIterator` trait yet and `TryFrom<I>` breaks blanket `TryFrom<T>` impl in std, so this is the next best thing - just a couple of inherent struct functions.

Related #142

### DIFF
--- a/src/array_string.rs
+++ b/src/array_string.rs
@@ -295,6 +295,36 @@ impl<const CAP: usize> ArrayString<CAP>
         Ok(())
     }
 
+    pub fn try_push_iterator<I: IntoIterator<Item = char>>(&mut self, value: I) -> Result<(), CapacityError<()>> {
+        let mut len = self.len();
+        let value = value.into_iter();
+        if value.size_hint().0 > self.capacity() - len {
+            return Err(CapacityError::new(()));
+        }
+
+        unsafe {
+            let mut ptr = self.as_mut_ptr().add(self.len());
+            for c in value {
+                let remaining_cap = self.capacity() - len;
+                match encode_utf8(c, ptr, remaining_cap) {
+                    Ok(n) => {
+                        len += n;
+                        ptr = ptr.add(n);
+                    }
+                    Err(_) => return Err(CapacityError::new(())),
+                }
+            }
+            self.set_len(len);
+            Ok(())
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


src/array_string.rs
```diff
@@ -295,6 +295,36 @@ impl<const CAP: usize> ArrayString<CAP>
         Ok(())
     }
 
+    pub fn try_push_iterator<I: IntoIterator<Item = char>>(&mut self, value: I) -> Result<(), CapacityError<()>> {
```
tbu-:
```suggestion
    pub fn try_extend<I: IntoIterator<Item = char>>(&mut self, value: I) -> Result<(), CapacityError<()>> {
```
I think the non-failing method is called `extend`: [`Vec::extend`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extend).

src/array_string.rs
```diff
@@ -295,6 +295,36 @@ impl<const CAP: usize> ArrayString<CAP>
...
+                let remaining_cap = self.capacity() - len;
+                match encode_utf8(c, ptr, remaining_cap) {
+                    Ok(n) => {
+                        len += n;
+                        ptr = ptr.add(n);
+                    }
+                    Err(_) => return Err(CapacityError::new(())),
+                }
+            }
+            self.set_len(len);
```
tbu-, line 317: Why is this implemented as `unsafe` instead of using `ArrayString::try_push`?

tbu- (commented):
I think the API additions look sane to me, even if they're not official `String` API.

src/array_string.rs
Pzixel: Because I think that this function should be as performant as possible, and calling `try_push` will modify the `len` on every iteration. This way we can set it once when we're done, which seems to be more efficient.

tbu-: Could you write a quick benchmark verifying that this improves performance over the naive `try_push` loop?

Pzixel: Sure, here are results on my machune:

```
test try_extend_bench ... bench:          28 ns/iter (+/- 2) = 146285 MB/s
test try_push_loop    ... bench:         336 ns/iter (+/- 13) = 12190 MB/s
```

I also fixed the other benches (at they were all optimized on my machine in just one LOAD of a capacity constant)

Pzixel:
Ping?

PR #296, opened 2025-03-28
feat+refactor: Bump MSRV to `1.57` and maximize `const`

Daniel-Aaron-Bloom:
Similar to #294, except it puts the changes behind a feature flag `const` and uses a macro `const_fn` to avoid code duplication. This allows the MSRV to stay low (still had to go up to `1.57` because I was getting weird `serde` errors, but if someone can figure those out, it might be possible to avoid the MSRV bump)


Daniel-Aaron-Bloom:
Ah, it's because `serde_test` upped its MSRV to `1.56` as of a couple versions ago. Also it's [deprecated and unmaintained](https://github.com/serde-deprecated/test).

Daniel-Aaron-Bloom:
@tbu- Do you think this an acceptable approach to address MSRV concerns?

PR #298, opened 2025-07-02
Implicitly use core, use alloc format for borsh environment

af-afk:
I had issues using this crate without std and the borsh flag enabled. This commit makes std explicitly opt in, and for the borsh feature flag uses alloc explicitly for format for error messages.


af-afk:
@bluss, I ran Cargo fmt without thinking about the consequences, let me know if I should reformat for this PR :) thanks for this crate!

PR #308, opened 2026-05-03
Fix double-free for ZSTs with Drop in .extend()

Shnatsel:
Zero-sized types that `impl Drop` get dropped twice in `.extend()`. This PR adds a test to demonstrate it, and fixes the issue.

This is a soundness bug, but likely not a big deal in practice, since this kind of type is really uncommon.

The issue was discovered by GPT-5.5 high


bluss:
Thanks for the perfect PR, that all looks valid.

I will have time to make a release next weekend

bluss:
fix released, sorry for the delay

PR #312, opened 2026-06-23
tests: skip two tests on 32-bit architectures

### LINKED ISSUE #311: v0.7.7 regression: tests fail on 32-bit architectures
It looks like https://github.com/bluss/arrayvec/pull/234 unintentionally regressed tests on 32-bit platforms where `usize` is (usually) the same size as `u32`. Removing the `panic!`ing code path for this case made the tests regress on 32-bit platforms since they no longer panic (logs from a build on i686-unknown-linux-gnu):

```
failures:

---- deny_max_capacity_arrayvec_value_const stdout ----
note: test did not panic as expected at tests/tests.rs:769:4

---- deny_max_capacity_arrayvec_value stdout ----
note: test did not panic as expected at tests/tests.rs:761:4

failures:
    deny_max_capacity_arrayvec_value
    deny_max_capacity_arrayvec_value_const
```

decathorpe:
Fixes #311

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -756,7 +756,7 @@ fn allow_max_capacity_arrayvec_type() {
 }
 
 #[should_panic(expected="largest supported capacity")]
-#[cfg(not(target_pointer_width = "16"))]
+#[cfg(not(any(target_pointer_width = "16", target_pointer_width = "32")))]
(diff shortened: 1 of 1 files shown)


bluss:
thanks!

